PR #18, opened 2017-10-20
User details claims

bdemers:
Spring Security does this when validating tokens remotely, so we need to set the user details when validating locally.

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
+++ b/oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
@@ -77,13 +77,17 @@ public class ConfigurableAccessTokenConverter extends DefaultAccessTokenConverte
 
     @Override
     public OAuth2Authentication extractAuthentication(Map<String, ?> map) {
+        // call super with the updated map
         OAuth2Authentication originalOAuth2Authentication = super.extractAuthentication(tweakScopeMap(map));
 
+        // If this token has UserAuthentication we need to rebuild it (we do not call setDetails() directly,
+        // as the implementation might change, currently it is a UsernamePasswordAuthenticationToken)
         Authentication originalUserAuthentication = originalOAuth2Authentication.getUserAuthentication();
         if (originalUserAuthentication != null) {
             UsernamePasswordAuthenticationToken newToken = new UsernamePasswordAuthenticationToken(originalUserAuthentication.getPrincipal(),
                     "N/A",
                     originalUserAuthentication.getAuthorities());
+            // now set the details on the new token directly, and return a new OAuth2Authentication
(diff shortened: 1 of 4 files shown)


oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
```diff
@@ -74,7 +77,16 @@ public OAuth2AccessToken extractAccessToken(String value, Map<String, ?> map) {
 
     @Override
     public OAuth2Authentication extractAuthentication(Map<String, ?> map) {
```
vijetmahabaleshwar-okta: Can you name "map" in a better way that suits this context?

oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
```diff
@@ -74,7 +77,16 @@ public OAuth2AccessToken extractAccessToken(String value, Map<String, ?> map) {
...
-        return super.extractAuthentication(tweakScopeMap(map));
-    }
+        OAuth2Authentication originalOAuth2Authentication = super.extractAuthentication(tweakScopeMap(map));
```
vijetmahabaleshwar-okta: Do you think we need some comments/javadoc for this?

oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
```diff
@@ -74,7 +77,16 @@ public OAuth2AccessToken extractAccessToken(String value, Map<String, ?> map) {
...
 
-}
+        Authentication originalUserAuthentication = originalOAuth2Authentication.getUserAuthentication();
+        if (originalUserAuthentication != null) {
+            UsernamePasswordAuthenticationToken newToken = new UsernamePasswordAuthenticationToken(originalUserAuthentication.getPrincipal(),
+                    "N/A",
+                    originalUserAuthentication.getAuthorities());
```
mraible: With JHipster, I had to grab the "groups" claim, create authorities, and set them accordingly. 

https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java#L104

oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
bdemers: This is the same name as the super class

oauth2/src/main/java/com/okta/spring/oauth/ConfigurableAccessTokenConverter.java
bdemers: 👍

PR #20, opened 2017-10-23
IDEA: Reuse ITs

bdemers:
Take a look at `integration-tests/oauth-mock-test-runner](integration-tests/oauth-mock-test-runner/README.md` for details.

This SHOULD NOT be merged in it's current state, it needs a little more discussion, but this PR should allow for demonstrating how this will work.

**NOTE:** this isn't exactly a black box as it was created with Spring's OAuth implementation in mind.


integration-tests/oauth-mock-test-runner/src/main/groovy/com/okta/test/mock/wiremock/HttpMock.groovy
```diff
@@ -0,0 +1,356 @@
...
+                .setHeader(Jwts.jwsHeader()
+                    .setKeyId('TEST_PUB_KEY_ID'))
+                .signWith(SignatureAlgorithm.RS256, keyPair.privateKey)
+                .compact()
+
+        // client_secret_basic auth scheme is used by default (Basic client_id:client_secret) by the clients
+        authHeader = "Basic " + "OOICU812:VERY_SECRET".bytes.encodeBase64().toString()
+    }
+
+    void configureHttpMock(WireMockServer wireMockServer) {
```
bdemers: I think we will need to abstract this configuration out of this class.  We need to decouple the configuration from the server.  I also think we should only load the stubs needed for the scenario and not all of them.  This may help us both detect errors, run tests in parallel, and allow for specific scenario checks (i.e. did the discovery URL get called)

integration-tests/oauth-mock-test-runner/src/main/groovy/com/okta/test/mock/application/ApplicationTestRunner.groovy
```diff
@@ -0,0 +1,140 @@
...
+
+    int doGetMockPort() {
+        return mockPort
+    }
+
+    @BeforeMethod
+    public void checkToRun(Method method) {
+        for (String disabledTest : scenario.disabledTests) {
+            if (method.getName().equals(disabledTest)) {
+                throw new SkipException("Skipping the disabled test - " + disabledTest);
```
bdemers: We shouldn't need this, we can configure which tests are run with TestNG

integration-tests/oauth-mock-test-runner/src/main/groovy/com/okta/test/mock/tests/CodeFlowLocalValidationIT.groovy
```diff
@@ -184,7 +184,7 @@ class CodeFlowLocalValidationIT extends ApplicationTestRunner {
             .statusCode(401)
     }
 
-    @Test
+    @Test (enabled = false)
```
bdemers: Why is this one disabled ?

integration-tests/oauth-mock-test-runner/src/main/groovy/com/okta/test/mock/application/ApplicationTestRunner.groovy
vijetmahabaleshwar-okta: This is provide each individual libraries the ability to disable tests easily.
We wouldn't want the client libraries to configure testng xml file for each scenario.
I think this is a good convenience for people who wouldn't know testng config and want to disable tests.

bdemers: This is fine for now, but we can probably handle this better in the future.   The testng.xml is only temporary, so we can either pass in the list of scenarios, or something.
We could also scan the yaml file and only run the scenarios that are defined.

This doesn't help us skip a specific test method, but I'm sure we can do something for that as well.

bdemers: Ideally we don't want to be skipping tests, but instead not running them. Any only skip the tests that _should_ be supported. (i.e. don't include auth code flow tests for the front end libs)

bdemers:
Replaced with #28

PR #32, opened 2017-12-20
Lazy loading properties configured via OIDC discovery and OktaPropertiesMappingEnvironmentPostProcessor

bdemers:
Now they work with spring-cloud-config!

Fixes: #26

This could sill use a few UTs in the `discovery` package, but any thoughts?  It feels like should be an easier way to do all of this.  Both the aliasing of the config properties and the injecting of the discovery metadata into the _standard_ spring properties.

There are no tests currently for spring-cloud-config, but I've manually verified that lazy loading the config properties allows for spring-cloud-config to be used to configure the `okta.*` props.

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
+++ b/oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
@@ -15,21 +15,21 @@
  */
 package com.okta.spring.oauth;
 
-import com.okta.spring.oauth.discovery.OidcDiscoveryClient;
-import com.okta.spring.oauth.discovery.OidcDiscoveryMetadata;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import com.okta.spring.oauth.discovery.DiscoveryPropertySource;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.boot.env.YamlPropertySourceLoader;
+import org.springframework.core.Ordered;
 import org.springframework.core.env.ConfigurableEnvironment;
-import org.springframework.core.env.EnumerablePropertySource;
 import org.springframework.core.env.Environment;
+import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.PropertySource;
 import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
-import org.springframework.web.client.ResourceAccessException;
+import org.springframework.util.StringUtils;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -81,31 +81,35 @@
...
-
     @Override
     public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
         environment.getPropertySources().addLast(remappedOktaToStandardOAuthPropertySource(environment));
-        environment.getPropertySources().addLast(discoveryPropertiesSource(environment));
-        environment.getPropertySources().addLast(loadYaml(new ClassPathResource("com/okta/spring/okta.yml")));
+        environment.getPropertySources().addLast(loadYaml(new FileSystemResource(new File(System.getProperty("user.home"), "okta/okta.yml")), false));
+        environment.getPropertySources().addLast(new DiscoveryPropertySource(environment));
+        environment.getPropertySources().addLast(oauthToClientPropertiesSource(environment));
+        environment.getPropertySources().addLast(loadYaml(new ClassPathResource("com/okta/spring/okta.yml"), true));
```
mraible: Will this override the one from `user.home`? Also, should the `user.home` path be `.okta/okta.yml`?

oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -81,31 +81,35 @@
...
-            throw new IllegalStateException("Failed to load yaml configuration from " + resource, ex);
+
+        if (resource.exists()) {
+            try {
+                return loader.load(resource.getFilename(), resource, null);
+            } catch (IOException ex) {
+                throw new IllegalStateException("Failed to load yaml configuration from " + resource, ex);
+            }
+        } else {
+            return new MapPropertySource("missing "+ resource.getFilename(), Collections.emptyMap());
```
mraible: Capitalize "missing"

oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -115,57 +119,53 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
...
+            this.environment = environment;
         }
 
         @Override
         public Object getProperty(String name) {
-            return getSource().get(name);
+            String orgUrlKey = "okta.client.orgUrl";
+            if (orgUrlKey.equals(name)) {
+                // first lookup the issuer
+                String issuerUrl = environment.getProperty(OKTA_OAUTH_PREFIX +"issuer");
```
mraible: Add space after `+`

oauth2/src/main/java/com/okta/spring/oauth/discovery/DiscoveryPropertySource.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public DiscoveryPropertySource(Environment environment) {
+        super("Okta-OIDC-Discovery-Client");
+        this.environment = environment;
+        this.isEnabled = !Boolean.parseBoolean(environment.getProperty(OKTA_OAUTH_DISCOVERY_DISABLED));
+    }
+
+    @Override
+    public Object getProperty(String name) {
+
```
mraible: Remove unnecessary line break.

oauth2/src/main/java/com/okta/spring/oauth/OktaPropertiesMappingEnvironmentPostProcessor.java
bdemers: It should be `.okta/okta.yml`.  First one wins, so after the standard spring property sources (sys props, env vars, etc) these will get applied (the last one, the ClassPathResource contains any defaults)

mraible: 👍

PR #45, opened 2018-03-12
Update the oidc-tck to 0.2.2

bdemers:
The example app needed to change to support this (this was back ported from the okta spring samples repo)

### DIFF
--- a/integration-tests/oauth2/src/test/resources/testRunner.yml
+++ b/integration-tests/oauth2/src/test/resources/testRunner.yml
@@ -35,7 +35,11 @@ scenarios:
     - --server.session.trackingModes=cookie
     - --okta.oauth2.localTokenValidation=false
 
+
   implicit-flow-local-validation:
+    disabledTests:
+    # Spring JWT parsing does NOT check the 'not-before' claim, (the 'exp' claim is validated)
+    - inactiveAccessTokenTest
     command: com.okta.spring.tests.oauth2.implicit.BasicImplicitFlowApplication
     args:
     - --server.port=${applicationPort}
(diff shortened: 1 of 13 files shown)


integration-tests/oauth2/src/test/resources/testRunner.yml
```diff
@@ -35,7 +35,11 @@ scenarios:
     - --server.session.trackingModes=cookie
     - --okta.oauth2.localTokenValidation=false
 
+
   implicit-flow-local-validation:
+    disabledTests:
+    # Spring JWT parsing does NOT check the 'not-before' claim, (the 'exp' claim is validated)
+    - inactiveAccessTokenTest
```
jmelberg-okta: Is there a prior version of Spring we could be using that is spec compliant?

bdemers: nope, but I didn't realize this claim is option https://tools.ietf.org/html/rfc7519#section-4.1.5

jmelberg-okta: It is optional to include this claim inside of a JWT - but it **must** be validated if it exists.

Is it possible to use another validation library? Seems [jose4j](https://bitbucket.org/b_c/jose4j/wiki/Home) is compliant.

bdemers: Use of the claim is optional, so they are still within spec. It isn't _easily_ possible to swap out the parser.  We can looking into adding extra validation into future releases.  But I'm really not too worried about this one just yet.  Okta doesn't set a not-before that is in the future

jmelberg-okta: Interesting.. it doesn't appear that we include that claim value in either token. We should probably remove [this test](https://github.com/okta/okta-oidc-tck/blob/c50c638bd284dc834db10c54f91d0ac498bee939/src/main/groovy/com/okta/test/mock/tests/ImplicitLocalValidationIT.groovy#L109) then.

PR #43, opened 2018-03-05
Add support for Spring Boot 2.0

mraible:
This is a proof-of-concept PR that shows it's possible to upgrade to Spring Boot 2.0. Fixes #30 and #39.

I've proven this works in https://github.com/oktadeveloper/okta-spring-boot-2-angular-5-example. However, one strange issue I ran into is:

```
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.0.0.RELEASE:run (default-cli) on project demo: 
An exception occurred while running. null: InvocationTargetException: Error creating bean with name 'oktaSdkClient' defined in class path resource [com/okta/spring/sdk/OktaSdkConfig.class]: 
Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.okta.sdk.client.Client]: 
Factory method 'oktaSdkClient' threw exception; nested exception is java.lang.IllegalArgumentException: baseUrl argument cannot be null. -> [Help 1]
```

Adding a `okta.client.orgUrl` property/value to my `application.properties` fixes this issue.

PR #58, opened 2018-05-15
Add support for Spring Boot 2.0

bdemers:
Based on the work of #43, but supports both Spring Boot 1 and 2

Fixes: #43

### DIFF
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -100,4 +100,22 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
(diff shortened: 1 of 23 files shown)


sdk/pom.xml
```diff
@@ -100,4 +100,22 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <!-- quick way to test for boo2 compatibility -->
```
mraible: boot2

PR #61, opened 2018-05-17
Cleanup duplicate spring version declaration

bdemers:
Exclude false positive from owasp dep scan

### DIFF
--- a/integration-tests/oauth2-boot2/pom.xml
+++ b/integration-tests/oauth2-boot2/pom.xml
@@ -29,7 +29,7 @@
     <properties>
         <test.mainClass>com.okta.spring.tests.oauth2.code.BasicRedirectCodeFlowApplication</test.mainClass>
 
-        <spring-boot.version>2.0.1.RELEASE</spring-boot.version>
+        <spring-boot.version>${spring-boot2.version}</spring-boot.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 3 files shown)


src/owasp/owasp-suppression.xml
```diff
@@ -50,4 +50,11 @@
        <cve>CVE-2017-6056</cve>
     </suppress>
 
+    <!-- wrong GAV for this CVE, it would be spring-security-oauth2 -->
```
mraible: would > should be?

integration-tests/oauth2-boot2/pom.xml
```diff
@@ -38,7 +38,7 @@
         <dependency>
             <groupId>org.springframework.security.oauth.boot</groupId>
             <artifactId>spring-security-oauth2-autoconfigure</artifactId>
-            <version>${spring-boot.version}</version>
+            <version>2.0.1.RELEASE</version>
```
mraible: Why do you use a variable on line 32, then hard-code it here? Also, should it be the latest (2.0.2)?

bdemers: This dep doesn't have the same release versions as the rest of spring boot, i should probably add a comment about that though (i'm also assuming it will get rolled into the bom at some point)

bdemers:
@mraible back to you

PR #73, opened 2018-06-28
Simplify okta oauth config aliases

### LINKED ISSUE #37: IntelliJ IDEA says clientId should be client-id and clientSecret should be client-secret
The bug is that if you use these values, no `clientId` is sent. If you use `clientId` and `clientSecret` it works. Screenshot of popup:

image

bdemers:
Fixes #37


bdemers:
Updating to Spring Security 5.1 soon, and all of this logic will change

PR #102, opened 2018-12-20
Update example for resource server

mraible:


mraible:
We should probably add an example showing WebFlux too.

PR #152, opened 2019-12-09
Add PKCE support

bdemers:
ITs to support PCKE have already been added to okta-oidc-tck (and will run automatically now that it is configured in testRunner.yml)

TL;DR: client-secret is no longer required


bdemers:
cc @mraible, @dogeared

Issue #136, opened 2019-07-23
Cannot set custom userInfoEndpoint user/oidc user service


bdemers:
This is caused by the use of a factory configured in `META-INF/spring.factories` 

```properties
org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer = \
  com.okta.spring.boot.oauth.OktaOAuth2Configurer
```

Maybe we can change this logic to lookup a bean from the context instead.  Otherwise, we might need to provide some other way to allow custom Authorities to be added to the current user.

bdemers:
(possibly getting the current context similar to this: https://github.com/okta/okta-spring-boot/blob/master/oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java#L46)

jkennedy4:
My use case for injecting my own code in the spring boot + oauth + okta flow is as follows.  We will be supporting 3 different types of applications on our side: a legacy (JBOSS) app with spring xml-config and custom "homegrown" security, a JBOSS app with spring security and a spring boot app.  We would like to add information to the OAuth2AuthenticationToken which is the object ultimately returned by Spring.  We would like to add our own authorities to the existing ones added by oauth/okta.  Also, we need to add a custom security user object to the http session which is used by the legacy app.  This information comes from our Oracle database, and we will use the loginId (from Okta) as a key to lookup the Oracle data.  That means our code HAS to follow the back-channel calls that are initiated from Spring's OidcAuthorizationCodeAuthenticationProvider.

msamm-r7:
i have a very similar use case to @jkennedy4 , where i want to add to the authorities during auth.

luupqvinh:
I have the same problem. I want to provide a custom oidc user service to add some custom authorities by querying a database.

bdemers:
Anyone watching this issue, I just created a PR with how I've been thinking about solving this. Feedback welcome!

TL;DR:

```java
@Bean
AuthoritiesProvider myCustomAuthoritiesProvider() {
    return (user, userRequest) -> lookupExtraAuthoritesByName(user.getAttributes().get("email"));
}
```

(There is still a little bit of cleanup to do on this PR, but I think it _should_ provide an easy way to add custom GrantedAuthorites)

dcods22:
Is there an update as to when this change would be in?

bdemers:
@dcods22 Can you take a look at #146 ? Would something like that work for you?

[...]

PR #154, opened 2020-01-15
Define and use OAuth2UserService beans

bdemers:
Previously we used an AbstractHttpConfigurer to set the Okta implementation of the OAuth2UserService (and oidc user service)
This made replacing Okta impl impossible to replace due to how/when those impls were configured.
Now they loaded automatically.

Fixes: #136


bdemers:
I'm not sure of the older versions of `HttpSecurity` allowed for this (the reactive equivalent always had), but 5.2 does 🎉

bdemers:
cc @mraible

PR #169, opened 2020-04-21
Typo in Spring placeholder syntax

helgp:
When resource server is located behind reverse proxy, redirect_uri parameter in the request, which the resource server sends to OKTA authorization server, contains invalid base URL.


bdemers:
Thanks @helgp! 
I'll try to add a quick test for this before we merge it

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -123,7 +123,7 @@ public Object getProperty(String name) {
 
     private PropertySource oktaRedirectUriPropertySource(Environment environment) {
         Map<String, Object> properties = new HashMap<>();
-        properties.put("spring.security.oauth2.client.registration.okta.redirect-uri", "{baseUrl}${okta.oauth2.redirect-uri}");
+        properties.put("spring.security.oauth2.client.registration.okta.redirect-uri", "${baseUrl}${okta.oauth2.redirect-uri}");
```
bdemers, line 126: I forgot about this one.  IIRC, the `{baseUrl }` is valid syntax and gets processed for each request
(so you can serve multiple domains from the same application)

Issue #147, opened 2019-11-08
Cannot load configuration class: com.okta.spring.boot.oauth.OktaOAuth2AutoConfig


d8tamgitadmin:
Can not start Spring Boot Application with the following error:

java.lang.IllegalStateException: Cannot load configuration class: com.okta.spring.boot.oauth.OktaOAuth2AutoConfig

`Caused by: java.lang.ClassNotFoundException: org.springframework.security.oauth2.client.oidc.web.logout.OidcClientInitiatedLogoutSuccessHandler
`

```
application.yaml:spring:
  profiles: dev
okta:
  oauth2:
    issuer: https://{}okta.com/oauth2/default
    client-id: foo
    client-secret: bar
```

pom.xml for oauth and security:
```
<dependency>
			<groupId>com.okta.spring</groupId>
			<artifactId>okta-spring-boot-starter</artifactId>
			<version>1.3.0</version>
		</dependency>
```

Java version is 1.8.

bdemers:
Which version of spring-bootstrap you using? 2.1?  
There was a minor class change incompatibility in spring security between boot 2.1 and 2.2.

If you are on Boot 2.1 use v1.2.x of this lib, otherwise use 1.3.x

I’ll make this more clear in the Readme, assuming that is the problem you are seeing.
Let me know!

jasonrberk:
please update the readme...this seems wrong:

> NOTE: This library works with Spring Boot 2.1+. If you need support for Spring Boot 1.5.x, use version version 0.6.

bdemers:
Hey @jasonrberk!
You are correct, I pushed a PR to get that updated.  Thanks for the heads up!

PR #172, opened 2020-05-22
Update minimum Spring Boot version in readme

bdemers:
from: #147


bdemers:
I would suggest updating, but you can also use Spring Security 2.1 directly.

Is something in Spring Boot 2.2 blocking you from upgrading?

PR #179, opened 2020-08-31
WIP - changes to support opaque token validation

arvindkrishnakumar-okta:


oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
```diff
@@ -71,6 +71,11 @@
      */
     private String postLogoutRedirectUri;
 
+    /**
+     * Identifies Opaque token
+     */
+    private Boolean isOpaque;
```
bdemers, line 77:
```suggestion
    private boolean opaque;
```

Use a simple boolean (only deal with two states) when possible

bdemers: don't prefix a bool with `is`

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -132,6 +132,9 @@ private PropertySource oktaStaticDiscoveryPropertySource(Environment environment
         Map<String, Object> properties = new HashMap<>();
         properties.put("spring.security.oauth2.resourceserver.jwt.issuer-uri", "${okta.oauth2.issuer}");
         properties.put("spring.security.oauth2.resourceserver.jwt.jwk-set-uri", "${okta.oauth2.issuer}/v1/keys");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.client-id", "${okta.oauth2.clientId}");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.client-secret", "${okta.oauth2.clientSecret}");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.introspection-uri", "${okta.oauth2.introspectionUri");
```
bdemers, line 137: These need to be conditional on `okta.oauth2.clientSecret` being set

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -75,6 +79,20 @@ public void init(HttpSecurity http) throws Exception {
                 } else {
                     log.debug("OAuth resource server not configured due to missing issuer property");
                 }
+
+                if (oktaOAuth2Properties.isOpaque() && !isEmpty(resourceServerProperties.getOpaquetoken().getIntrospectionUri())) {
+                    OpaqueTokenIntrospector introspectionClient = new NimbusOpaqueTokenIntrospector(
+                        resourceServerProperties.getOpaquetoken().getIntrospectionUri(),
+                        oktaOAuth2Properties.getClientId(),
+                        oktaOAuth2Properties.getClientSecret());
```
bdemers, line 87: Tthis logic _should_ be available in the context: `context.getBean(OpaqueTokenIntrospector.class)`

Care needs to be taken here so both `http.oauth2ResourceServer().jwt()` and `http.oauth2ResourceServer().opaqueToken()` are both not called as this WILL create a problem.

It maybe possible to inject the bits we need at different level now (I haven't looked in a while)

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -75,6 +79,20 @@ public void init(HttpSecurity http) throws Exception {
...
+
+                    OpaqueTokenAuthenticationProvider opaqueTokenAuthenticationProvider =
+                        new OpaqueTokenAuthenticationProvider(introspectionClient);
+
+                    http.oauth2ResourceServer().opaqueToken()
+                        .authenticationManager(authentication -> opaqueTokenAuthenticationProvider
+                            .authenticate(SecurityContextHolder.getContext().getAuthentication()));
```
bdemers, line 94: We would also need to deal with the group claim converter similar to line 98/116 below

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
arvindkrishnakumar-okta: Sorry, did you mean conditional on `okta.oauth2.opaque` being set instead?

bdemers: No, assuming there is no downside to configuring both. I'm just not sure what happens when you set the introspection-uri and ID but no secret is used (I assume that would fail).
The `okta.oauth2.opauqe` flag would just trigger the default SecurityConfigurer to use opaque or jwt (currently the default is to always use JWT)

NOTE, there could be other conditionals here that I'm not thinking about, so we need to make sure we have good testing around this

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospector.java#L182-L217 does exactly this. 

We can make use of this introspector:

```
OpaqueTokenIntrospector opaqueTokenIntrospector = context.getBean(OpaqueTokenIntrospector.class);
http.oauth2ResourceServer().opaqueToken().introspector(opaqueTokenIntrospector);
```

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: Should we deal with separating these flows at `runtime` or during `init`?

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -132,6 +134,14 @@ private PropertySource oktaStaticDiscoveryPropertySource(Environment environment
         Map<String, Object> properties = new HashMap<>();
         properties.put("spring.security.oauth2.resourceserver.jwt.issuer-uri", "${okta.oauth2.issuer}");
         properties.put("spring.security.oauth2.resourceserver.jwt.jwk-set-uri", "${okta.oauth2.issuer}/v1/keys");
+
+        if (BooleanUtils.toBoolean("${okta.oauth2.opaque}") &&
```
arvindkrishnakumar-okta, line 138: TODO: When we have both jwt and opaque, the default should be JWT.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: addressed this.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: addressed this.

PR #185, opened 2020-09-10
Add Opaque Token Validation Support for Servlet Web Application Type

arvindkrishnakumar-okta:
### Issue:

[OKTA-323350](https://oktainc.atlassian.net/browse/OKTA-323350)

### Expected behavior implemented in this PR:

| Issuer/Org URL | Client Specified Config | Resource Server Start Mode|
| ------------- | ------------- |---------------------------------------|
|     Root         |         Opaque                 |        Opaque |
|     Root         |            Jwt                     |   Server won't start |
|     Root         |            Unset                     |   Opaque |
|  Non-root     |            Jwt                     |          Jwt |
|  Non-root     |         Opaque                 |       Opaque |
|  Non-root     |         Unset                 |       Neither Jwt nor Opaque |

ℹ️ **Note:** When the server is configured/set to start in Opaque Token mode, if the property `okta.oauth2.client-secret` is not set (empty), then the startup would fail with error and this is the desired behavior. This is because, we will need the client-secret at runtime to hit the introspection endpoint to validate the Opaque token and we do not want the user to see a 401 error at that stage with a missing client-secret. To prevent that, we are validating that before the server startup and would not let the server to start in such a case.

**Note:** This covers only Servlet web application type (does not cover "Reactive/Webflux" applications).


oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -110,4 +128,22 @@ private void configureResourceServer(HttpSecurity http, OktaOAuth2Properties okt
...
+                                                    OAuth2ResourceServerProperties resourceServerProperties) {
+        return oktaOAuth2Properties.isOpaque()
+            && !isEmpty(resourceServerProperties.getOpaquetoken().getClientId())
+            && !isEmpty(resourceServerProperties.getOpaquetoken().getClientSecret())
+            && !isEmpty(resourceServerProperties.getOpaquetoken().getIntrospectionUri());
+    }
+
+    private boolean isRootOrgIssuer(String issuerUri) {
+        try {
+            ConfigurationValidator.assertOrgUrl(issuerUri);
```
bdemers, line 142: Fine for debugging, but this has the potential to fail in other was, for example, if the issuer URL was non-https.

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
```diff
@@ -71,6 +71,11 @@
      */
     private String postLogoutRedirectUri;
 
+    /**
+     * Identifies Opaque token
+     */
+    private Boolean opaque = Boolean.FALSE;
```
bdemers, line 77: Use small b boolean, otherwise, you need to deal with 3 states

oauth2/src/main/resources/application.properties
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+logging.level.org.springframework.security.web=INFO
+logging.level.org.springframework.web.client.RestTemplate=DEBUG
+logging.level.com.okta.spring.boot.oauth=DEBUG
```
bdemers, line 19: Remove this file

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
```diff
@@ -65,4 +73,21 @@ OidcClientInitiatedLogoutSuccessHandler oidcLogoutSuccessHandler(OktaOAuth2Prope
...
+
+    @Bean
+    @ConditionalOnProperty(name="okta.oauth2.opaque", havingValue="true")
+    OpaqueTokenIntrospector oktaOpaqueTokenIntrospector(OktaOAuth2Properties oktaOAuth2Properties,
+                                                        OAuth2ResourceServerProperties oAuth2ResourceServerProperties) {
+        RestTemplate restTemplate = new RestTemplate(Arrays.asList(
+            new FormHttpMessageConverter(),
+            new StringHttpMessageConverter()));
+        restTemplate.setErrorHandler(new OAuth2ErrorResponseErrorHandler());
+        restTemplate.getInterceptors().add(new UserAgentRequestInterceptor());
```
bdemers, line 85: It's probably worth pulling his logic out into a common method as we have a few similar usages already.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOpaqueTokenIntrospector.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Okta's implementation of Spring security's interface {@link OpaqueTokenIntrospector} based on
+ * {@link org.springframework.security.oauth2.server.resource.introspection.NimbusOpaqueTokenIntrospector}.
+ */
+public class OktaOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
```
bdemers, line 56: There seems to be a lot of logic in this class.  Why not use `NimbusOpaqueTokenIntrospector` instead?

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
```diff
@@ -71,6 +71,11 @@
      */
     private String postLogoutRedirectUri;
 
+    /**
+     * Identifies Opaque token
```
bdemers, line 75: Make sure this is in proper Javadoc format, and more verbose. As this text is used for the generated complication used in IDEs

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
```diff
@@ -154,6 +159,18 @@ public void setPostLogoutRedirectUri(String postLogoutRedirectUri) {
         this.postLogoutRedirectUri = postLogoutRedirectUri;
     }
 
+    public boolean getOpaque() {
```
bdemers, line 162: remove this in favor of `isOpaque()`

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -132,6 +144,9 @@ private PropertySource oktaStaticDiscoveryPropertySource(Environment environment
         Map<String, Object> properties = new HashMap<>();
         properties.put("spring.security.oauth2.resourceserver.jwt.issuer-uri", "${okta.oauth2.issuer}");
         properties.put("spring.security.oauth2.resourceserver.jwt.jwk-set-uri", "${okta.oauth2.issuer}/v1/keys");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.client-id", "${okta.oauth2.clientId}");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.client-secret", "${okta.oauth2.clientSecret}");
+        properties.put("spring.security.oauth2.resourceserver.opaquetoken.introspection-uri", "${okta.oauth2.issuer}/v1/introspect");
```
bdemers, line 149: What happens when the `client-id` is set but not `client-secret` ?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOpaqueTokenIntrospector.java
arvindkrishnakumar-okta: This is because, we need to customise the request body and headers to send additional params to Okta's introspection endpoint. The `NimbusOpaqueTokenIntrospector` just sends the token to the introspection endpoint and this wont suffice. Hence the customization.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
arvindkrishnakumar-okta: resource server will start fine but the introspection wont work as it needs the secret to be sent.

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
arvindkrishnakumar-okta: done

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -92,10 +104,16 @@ private void configureLogin(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Pr
...
+
+        http.oauth2ResourceServer()
+            .jwt().jwtAuthenticationConverter(new OktaJwtAuthenticationConverter(oktaOAuth2Properties.getGroupsClaim()));
+    }
+
+    private void configureResourceServerWithOpaqueTokenValidation(HttpSecurity http, ApplicationContext context) throws Exception {
 
         http.oauth2ResourceServer()
-                .jwt().jwtAuthenticationConverter(new OktaJwtAuthenticationConverter(oktaOAuth2Properties.getGroupsClaim()));
+            .opaqueToken().introspector(context.getBean(OpaqueTokenIntrospector.class));
```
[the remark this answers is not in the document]
arvindkrishnakumar-okta: done

oauth2/src/main/resources/application.properties
arvindkrishnakumar-okta: will do, kept it in place temporarily for debugging purposes.

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
arvindkrishnakumar-okta: will do.

oauth2/src/main/resources/application.properties
arvindkrishnakumar-okta: removed.

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
arvindkrishnakumar-okta: done

oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java
arvindkrishnakumar-okta: Updated to:

` /**
     * Signals need for opaque token validation/introspection (if set to true).
     */
`

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: a non-https issuer URL will yield a `false` return value from this method (https check is handled [here](https://github.com/okta/okta-commons-java/blob/master/config-check/src/main/java/com/okta/commons/configcheck/ConfigurationValidator.java#L251) by the library).

bdemers: Sorry, right bad example. but detecting if a URL is a root issuer or not shouldn't be done this way as that isn't the intent of this method.  My suggestion is to check if the URL contains something like `/oauth/v1` (or something), if not, it's a root org issuer.  This also prevents using the exception handling for the normal expected case.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
arvindkrishnakumar-okta: done

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOpaqueTokenIntrospector.java
bdemers: I'm not sure I'm following.
You can just set the auth headers in the `RestTemplate`

https://github.com/spring-projects/spring-security/blob/bf067d679fb5990d6b3d31cee292af171f8a5014/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospector.java#L82-L83

The client-id and secret are not needed when you configure the RestTemplate directly

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
bdemers: ok, so we need to make sure that runtime error does NOT occur.

We should probably skip the `isOpaque()` flag and just configure what is available, in cases where both are available the developer can use the `.jwt()` and `.opaqueToken()` methods, and we are not doing anything special or confusing.

Similaraly in the case of a "root/org" issuer, the JWT properties should NOT be configured.  This will alow the default Spring configuration to automatically use opaque tokens as that is the only one available.

In cases where a user tries to explicitly use `jwt()` in cases where it's not configured, it will fail at startup (which is ideal).

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: So if we detect root org issuer, we should enforce Opaque Token introspection flow, right? If so,  how should we setup the introspection URI? Presently it is set to a value relative to the issuer URL (`${okta.oauth2.issuer}/v1/introspect`) - is that right? or should we let the user supply the introspection uri?.

bdemers: Yes, making it relative to the issuer URL.  The user should NOT need to supply the introspection URL manually.  In this case, we can hard code it because it's not going to change.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOpaqueTokenIntrospector.java
arvindkrishnakumar-okta: got it, updated.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
arvindkrishnakumar-okta: makes sense.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -110,4 +118,8 @@ private void configureResourceServer(HttpSecurity http, OktaOAuth2Properties okt
 
         return accessTokenResponseClient;
     }
+
+    private boolean isRootOrgIssuer(String issuerUri) {
+        return !issuerUri.contains("/oauth2/default");
```
bdemers, line 123: don't use `default` that is actually an ID,  and could be any string

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -56,4 +58,14 @@ private RestOperations restOperations() {
...
     }
+
+    @Bean
+    @ConditionalOnMissingBean
+    OpaqueTokenIntrospector opaqueTokenIntrospector(OktaOAuth2Properties oktaOAuth2Properties,
+                                                    OAuth2ResourceServerProperties oAuth2ResourceServerProperties) {
+        return new NimbusOpaqueTokenIntrospector(
+            oAuth2ResourceServerProperties.getOpaquetoken().getIntrospectionUri(),
+            oktaOAuth2Properties.getClientId(),
+            oktaOAuth2Properties.getClientSecret());
```
bdemers, line 69: Where is the RestTemplate set?

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
```diff
@@ -155,7 +150,8 @@ class AutoConfigConditionalTest implements HttpMock {
 
         // with properties it loads correctly
         webContextRunner().withPropertyValues(
-                "okta.oauth2.issuer=https://test.example.com/")
+                "okta.oauth2.issuer=https://test.example.com/",
+        "okta.oauth2.client-id=test-client-id", "okta.oauth2.client-secret=test-client-secret")
```
bdemers, line 154: The existing tets _should_ work as is, add new tests to test other permutations

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: fixed

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
arvindkrishnakumar-okta: noted.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: addressed

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
arvindkrishnakumar-okta: @bdemers I've pushed updates to code that will deal with the possible cases listed in the description section of this PR. Please let me know if I missed any use case.

briandemers-okta:
I love the table! We also need rows for "Client Specified Config" == unset for each option

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2017 Okta, Inc.
```
briandemers-okta, line 2: I'm guessing this should be 202

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
```diff
@@ -0,0 +1,58 @@
...
+            .contentType(ContentType.ANY)
+            .redirects()
+                .follow(false)
+        .when()
+            .get("http://localhost:${applicationPort}/everyone")
+        .then()
+            .statusCode(401)
+            .header("WWW-Authenticate", startsWith("Bearer"))
+            .body(Matchers.equalTo(ERROR_401))
+    }
```
briandemers-okta, line 57: you may want to move this test to: https://github.com/okta/okta-oidc-tck/blob/master/src/main/groovy/com/okta/test/mock/tests/ImplicitRemoteValidationIT.groovy

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -62,21 +66,41 @@ public void init(HttpSecurity http) throws Exception {
...
-                    configureResourceServer(http, oktaOAuth2Properties);
-                } else {
-                    log.debug("OAuth resource server not configured due to missing issuer property");
+                OAuth2ResourceServerConfigurer oAuth2ResourceServerConfigurer =
+                    http.getConfigurer(OAuth2ResourceServerConfigurer.class);
+
+                if (oAuth2ResourceServerConfigurer != null) {
+
+                    // code below is wrapped in AccessController to address findbugs error 'DP_DO_INSIDE_DO_PRIVILEGED'
+                    //noinspection unchecked
```
briandemers-okta, line 81: why `noinspection` ?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -62,21 +66,41 @@ public void init(HttpSecurity http) throws Exception {
...
+                    Field jwtConfigurerField = (Field) AccessController.doPrivileged((PrivilegedAction) () -> {
+                        Field result = null;
+                        try {
+                            result = OAuth2ResourceServerConfigurer.class.getDeclaredField("jwtConfigurer");
+                            result.setAccessible(true);
+                        } catch (NoSuchFieldException e) {
+                            log.error("Could not get field 'jwtConfigurer' of {} via reflection",
```
briandemers-okta, line 88: nit: if you are continuing is this a warning?

Either way, tell the developer what to do in this case. Maybe mention that this version of Spring Security is not compatible?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -92,16 +116,15 @@ private void configureLogin(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Pr
         }
     }
 
-    private void configureResourceServer(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Properties) throws Exception {
+    private void configureResourceServerForOpaqueTokenValidation(HttpSecurity http) throws Exception {
```
briandemers-okta, line 119: Where is this called for non-root issuers?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -92,16 +116,15 @@ private void configureLogin(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Pr
...
-        http.oauth2ResourceServer()
-                .jwt().jwtAuthenticationConverter(new OktaJwtAuthenticationConverter(oktaOAuth2Properties.getGroupsClaim()));
+        http.oauth2ResourceServer().opaqueToken();
     }
 
     private OAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantRequest> accessTokenResponseClient() {
 
         RestTemplate restTemplate = new RestTemplate(Arrays.asList(new FormHttpMessageConverter(),
-                                                                   new OAuth2AccessTokenResponseHttpMessageConverter()));
+            new OAuth2AccessTokenResponseHttpMessageConverter()));
```
briandemers-okta, line 127: revert this white space change

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -45,15 +58,46 @@
...
-            decoder.setJwtValidator(TokenUtil.jwtValidator(oAuth2ResourceServerProperties.getJwt().getIssuerUri(), oktaOAuth2Properties.getAudience()));
-            decoder.setRestOperations(restOperations());
-            return decoder;
+        NimbusJwtDecoderJwkSupport decoder = new NimbusJwtDecoderJwkSupport(oAuth2ResourceServerProperties.getJwt().getJwkSetUri());
+        decoder.setJwtValidator(TokenUtil.jwtValidator(oAuth2ResourceServerProperties.getJwt().getIssuerUri(), oktaOAuth2Properties.getAudience()));
+        decoder.setRestOperations(restOperations());
+        return decoder;
     }
 
+
```
briandemers-okta, line 67:
```suggestion
```

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -45,15 +58,46 @@
...
+            OAuth2AuthenticatedPrincipal principal = delegate.introspect(token);
+            Collection<GrantedAuthority> mappedAuthorities =
+                (Collection<GrantedAuthority>) TokenUtil.tokenClaimsToAuthorities(principal.getAttributes(), oktaOAuth2Properties.getGroupsClaim());
+            return new DefaultOAuth2AuthenticatedPrincipal(
+                principal.getName(), principal.getAttributes(), mappedAuthorities);
+        };
+    }
+
+    @Bean
+    RestTemplate restTemplate() {
```
briandemers-okta, line 94: Creating a `RestTemplate` bean like this may conflict with other applications.  Where is this one used?  Maybe just create it an instance for that case

oauth2/src/main/java/com/okta/spring/boot/oauth/TokenUtil.java
```diff
@@ -90,4 +93,35 @@ private TokenUtil(){}
...
+    static boolean isRootOrgIssuer(String issuerUri) throws MalformedURLException {
+        String uriPath = new URL(issuerUri).getPath();
+
+        if (Strings.hasText(uriPath)) {
+            String[] tokenizedUri = uriPath.substring(uriPath.indexOf("/")+1).split("/");
+
+            if (tokenizedUri.length >= 2 &&
+                "oauth2".equals(tokenizedUri[0]) &&
+                Strings.hasText(tokenizedUri[1])) {
+                log.debug("{} is not root/org issuer", issuerUri);
```
briandemers-okta, line 119:
```suggestion
                log.debug("{} is not an Okta root/org issuer", issuerUri);
```

oauth2/src/main/java/com/okta/spring/boot/oauth/TokenUtil.java
```diff
@@ -90,4 +93,35 @@ private TokenUtil(){}
...
+                return false;
+            }
+        }
+
+        log.debug("{} is root/org issuer", issuerUri);
```
briandemers-okta, line 124:
```suggestion
        log.debug("{} is an Okta root/org issuer, JWT or opaque token validation may be enabled", issuerUri);
```

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -132,6 +144,9 @@ private PropertySource oktaStaticDiscoveryPropertySource(Environment environment
         Map<String, Object> properties = new HashMap<>();
         properties.put("spring.security.oauth2.resourceserver.jwt.issuer-uri", "${okta.oauth2.issuer}");
         properties.put("spring.security.oauth2.resourceserver.jwt.jwk-set-uri", "${okta.oauth2.issuer}/v1/keys");
+        properties.put("spring.security.oauth2.resourceserver.opaque-token.client-id", "${okta.oauth2.client-id}");
+        properties.put("spring.security.oauth2.resourceserver.opaque-token.client-secret", "${okta.oauth2.client-secret}");
+        properties.put("spring.security.oauth2.resourceserver.opaque-token.introspection-uri", "${okta.oauth2.issuer}/v1/introspect");
```
briandemers-okta, line 149: What happens here when `client-secret` is NOT set?

oauth2/src/test/groovy/com/okta/spring/boot/oauth/TokenUtilTest.groovy
```diff
@@ -106,4 +106,17 @@ class TokenUtilTest {
                                                                                                     new SimpleGrantedAuthority("sEa"))).and(
                                                                                                 hasSize(3))
     }
+
+    @Test
+    void issuerUri_rootOrgTest() {
+        assertThat "issuer uri expected to be root/org)", TokenUtil.isRootOrgIssuer("https://sample.okta.com")
+        assertThat "issuer uri expected to be root/org)", TokenUtil.isRootOrgIssuer("https://dev-12345.oktapreview.com/")
+        assertThat "issuer uri expected to be root/org)", TokenUtil.isRootOrgIssuer("https://example.io")
```
briandemers-okta, line 114:
```suggestion
        assertThat "issuer uri expected to be root/org", TokenUtil.isRootOrgIssuer("https://sample.okta.com")
        assertThat "issuer uri expected to be root/org", TokenUtil.isRootOrgIssuer("https://dev-12345.oktapreview.com/")
        assertThat "issuer uri expected to be root/org", TokenUtil.isRootOrgIssuer("https://example.io")
```

oauth2/src/test/groovy/com/okta/spring/boot/oauth/TokenUtilTest.groovy
```diff
@@ -106,4 +106,17 @@ class TokenUtilTest {
...
+    }
+
+    @Test
+    void issuerUri_nonRootOrgTest() {
+        assertThat "issuer uri expected to be non-root/org)", !TokenUtil.isRootOrgIssuer("https://sample.okta.com/oauth2/default")
+        assertThat "issuer uri expected to be non-root/org)", !TokenUtil.isRootOrgIssuer("https://example.io/oauth2/ausvd5ple5TRRsbcJ0h7")
```
briandemers-okta, line 120:
```suggestion
        assertThat "issuer uri expected to be non-root/org", !TokenUtil.isRootOrgIssuer("https://sample.okta.com/oauth2/default")
        assertThat "issuer uri expected to be non-root/org", !TokenUtil.isRootOrgIssuer("https://example.io/oauth2/ausvd5ple5TRRsbcJ0h7")
```

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: That was IntelliJ's recommendation to suppress the `unchecked` warnings in the below code segment.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: This method will not be called for `non-root` issuer case. Because, for a `non-root` issuer case, the flow could take either `jwt` or `Opaque` token flow depending on what the user has provisioned/configured.

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
arvindkrishnakumar-okta: updated

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: fixed

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: `RestTemplate` bean type is scanned for by `NimbusOpaqueTokenInstrospector` because of the need to make a network call to remote introspection endpoint. I've put in a conditional on this bean to address your concern/comment.

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
arvindkrishnakumar-okta: done, see https://github.com/okta/okta-oidc-tck/pull/85.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: removed it safely (unused).

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: updated with an apt logger message.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: it will be invoked after we inspect the jwtconfigurer and determine that jwt is not configured.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
arvindkrishnakumar-okta: presently it starts the server and will fail in introspection step with a 401 unauthorized error. I will change this to "server should not start" behavior.

arvindkrishnakumar-okta: addressed this.

README.md
```diff
@@ -222,6 +259,28 @@ You'll be redirected automatically to an Okta login page. Once you successfully
...
+
+```java
+System.setProperty("https.proxyHost", "https://example-proxy.com");
+System.setProperty("https.proxyPort", "443");
+```
+
+See [here](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) for the complete list of properties.
+
+**Note:**  Spring WebFlux (and `WebClient`) do not support these properties. (See [spring-projects/spring-security#8882](https://github.com/spring-projects/spring-security/issues/8882)).
+
```
bdemers, line 283: Can you move these out to another PR, we can probably merge this separately.
Just make sure you tested against a proxy before merging into the readme (NOTE: IIRC `localhost` is in the default non-proxy-host list)

README.md
```diff
@@ -233,7 +292,13 @@ To integrate the [Okta Java SDK](https://github.com/okta/okta-sdk-java) into you
...
 ```
 
-Then define the `okta.client.token` property. See [creating an API token](https://developer.okta.com/docs/api/getting_started/getting_a_token) for more info.
+Then define the following properties:
+
+| Key | Description |
+------|--------------
+| `okta.client.orgUrl` | Your Okta Url: `https://{yourOktaDomain}`, i.e. `https://dev-123456.okta.com` |
+| `okta.client.token` | An Okta API token, see [creating an API token](https://developer.okta.com/docs/api/getting_started/getting_a_token) for more info. |
+
```
bdemers, line 301: Looks like you need to merge `master` into `release_v2` ?

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
```diff
@@ -0,0 +1,43 @@
...
+    @Test
+    void groupAccessTest() {
+        given()
+            .header("Authorization", "Bearer ${IMPLICIT_FLOW_REMOTE_VALIDATION.definition.accessTokenJwt}")
+            .redirects()
+                .follow(false)
+        .when()
+            .get("http://localhost:${applicationPort}/everyone")
+        .then()
+            .body(Matchers.equalTo("Everyone has Access: joe.coder@example.com"))
```
bdemers, line 40: add a check for the 200 status code too

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -62,21 +66,49 @@ public void init(HttpSecurity http) throws Exception {
...
+                        }
+                        return result;
+                    });
+
+                    if (jwtConfigurerField != null) {
+                        OAuth2ResourceServerConfigurer.JwtConfigurer jwtConfigurerValue =
+                            (OAuth2ResourceServerConfigurer.JwtConfigurer) jwtConfigurerField.get(oAuth2ResourceServerConfigurer);
+
+                        if (jwtConfigurerValue != null) {
+                            log.info("JWT configurer is set in OAuth resource server configuration. " +
```
bdemers, line 99: nit: set these loggers to `debug`

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -62,21 +66,49 @@ public void init(HttpSecurity http) throws Exception {
...
+                                "JWT validation will be configured.");
+                            http.oauth2ResourceServer()
+                                .jwt().jwtAuthenticationConverter(new OktaJwtAuthenticationConverter(oktaOAuth2Properties.getGroupsClaim()));
+                        } else {
+                            log.info("JWT configurer is NOT set in OAuth resource server configuration. " +
+                                "Opaque Token validation/introspection will be configured.");
+                            configureResourceServerForOpaqueTokenValidation(http, oktaOAuth2Properties);
+                        }
+                    } else {
+                        log.warn("JWT configurer field was not found in OAuth resource server configuration");
```
bdemers, line 109: nit: Add something about checking this project for updates, https://github.com/okta/okta-spring-boot

Thinking about this again 🤔 we should probably fail in this case (still with a note about checking the project page)

My thoughts around this are our JWT checking validates more than the out of the box Spring Sec config.
Defaulting to a weaker security stance could raise other issues.  So just fail, with a nice long error message about the compatibility between versions, and checking for updates on the GitHub project's page.

A user should ONLY see this type of message when updating one of the versions without the other, so failing fast  is probably the best bet

(sorry for going back and forth on this)

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -45,15 +53,35 @@
...
+
+    @Bean
+    @Conditional(OktaOpaqueTokenIntrospectConditional.class)
+    OpaqueTokenIntrospector opaqueTokenIntrospector(OktaOAuth2Properties oktaOAuth2Properties,
+                                                    OAuth2ResourceServerProperties oAuth2ResourceServerProperties) {
+
+        OpaqueTokenIntrospector delegate = new NimbusOpaqueTokenIntrospector(
+            oAuth2ResourceServerProperties.getOpaquetoken().getIntrospectionUri(),
+            oAuth2ResourceServerProperties.getOpaquetoken().getClientId(),
+            oAuth2ResourceServerProperties.getOpaquetoken().getClientSecret());
```
bdemers, line 77: Can you set the RestTemplate here?

bdemers: Or is this object used for WebFlux too?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOpaqueTokenIntrospectConditional.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.spring.boot.oauth;
+
+import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+
+public class OktaOpaqueTokenIntrospectConditional extends AllNestedConditions {
```
bdemers, line 21: can this be package protected?

oauth2/src/main/java/com/okta/spring/boot/oauth/TokenUtil.java
bdemers, line 124:
```suggestion
        log.debug("The issuer URL: '{}' is an Okta root/org authorization server", issuerUri);
```

oauth2/src/main/java/com/okta/spring/boot/oauth/TokenUtil.java
bdemers, line 119:
```suggestion
                log.debug("The issuer URL: '{}' is an Okta custom authorization server", issuerUri);
                
```

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
```diff
@@ -265,7 +261,8 @@ class AutoConfigConditionalTest implements HttpMock {
         webContextRunner().withPropertyValues(
                 "okta.oauth2.issuer=https://test.example.com/",
                 "spring.security.oauth2.client.provider.okta.issuerUri=${mockBaseUrl()}", // work around to not validate the https url
-                "okta.oauth2.client-id=test-client-id")
+                "okta.oauth2.client-id=test-client-id",
+                "okta.oauth2.client-secret=test-client-secret")
```
bdemers, line 265: Why was the secret added here?

bdemers (changes requested):
Looks pretty good!
Just a few nits, left :D

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
arvindkrishnakumar-okta: Because the issuer used here is root org. Therefore, Opaque flow is chosen and this requires `client-secret` to be set otherwise the test will fail.

README.md
arvindkrishnakumar-okta: Yes, I merged the master into this feature branch and that is why these changes show up here. Will revert that. Once this PR is merged, I'll create a separate PR to merge `release_v2` into master.

README.md
arvindkrishnakumar-okta: Ref my comment above.

integration-tests/oauth2/src/test/groovy/com/okta/spring/tests/oauth2/implicit/ImplicitRemoteValidationGroupIT.groovy
arvindkrishnakumar-okta: will do.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: will do.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
arvindkrishnakumar-okta: will update.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: Actually, there is no real need for us to set the `RestTemplate` here (but we could, if we'd like that customization here). `NimbusOpaqueTokenIntrospector` will new a `RestTemplate` for us if we don't set it here. This is being used for Webflux too.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest.groovy
```diff
@@ -54,6 +57,9 @@ class OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest {
         assertThat environment.getProperty(TOKEN_URI), is("https://issuer.example.com/foobar/v1/token")
         assertThat environment.getProperty(USER_INFO_URI), is("https://issuer.example.com/foobar/v1/userinfo")
         assertThat environment.getProperty(PROVIDER_KEYS_URI), is("https://issuer.example.com/foobar/v1/keys")
+        assertThat environment.getProperty(RS_CLIENT_ID), is("test-client-id")
+        assertThat environment.getProperty(RS_CLIENT_SECRET), is("test-client-secret")
+        assertThat environment.getProperty(RS_INTROSPECTION_URI), is("https://issuer.example.com/foobar/v1/introspect")
```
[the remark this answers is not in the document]
arvindkrishnakumar-okta: done, pushed updates.

arvindkrishnakumar-okta:
> Looks pretty good!
> Just a few nits, left :D

Pushed updates, ready for review again.

README.md
arvindkrishnakumar-okta: Done

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: As discussed earlier today, pushed an update that would add `UserAgentRequestInterceptor` header to the `RestTemplate` injected by `NimbusOpaqueTokenIntrospector`.

bdemers: Are you still switching to the other constructor?

arvindkrishnakumar-okta: If I switch to the other constructor (`NimbusOpaqueTokenIntrospector(String introspectionUri, RestOperations restOperations)`, I would still have to find a way to inject client-id and client-secret and then stub a `        restTemplate.getInterceptors().add(new BasicAuthenticationInterceptor(clientId, clientSecret))` in. Either way, it should be fine and I chose the other way.

arvindkrishnakumar-okta: updated now.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -45,15 +54,39 @@
...
         restTemplate.getInterceptors().add(new UserAgentRequestInterceptor());
         return restTemplate;
     }
+
+    @Bean
+    @Conditional(OktaOpaqueTokenIntrospectConditional.class)
+    OpaqueTokenIntrospector opaqueTokenIntrospector(OktaOAuth2Properties oktaOAuth2Properties,
+                                                    OAuth2ResourceServerProperties oAuth2ResourceServerProperties) {
+
+        RestTemplate restTemplate = (RestTemplate) restOperations();
```
bdemers, line 75: nit: avoid casting here

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
```diff
@@ -45,15 +54,39 @@
...
+            oAuth2ResourceServerProperties.getOpaquetoken().getClientSecret()));
+
+        OpaqueTokenIntrospector delegate = new NimbusOpaqueTokenIntrospector(
+            oAuth2ResourceServerProperties.getOpaquetoken().getIntrospectionUri(),
+            restTemplate);
+
+        return token -> {
+            OAuth2AuthenticatedPrincipal principal = delegate.introspect(token);
+            Collection<GrantedAuthority> mappedAuthorities =
+                (Collection<GrantedAuthority>) TokenUtil.tokenClaimsToAuthorities(principal.getAttributes(), oktaOAuth2Properties.getGroupsClaim());
```
bdemers, line 87:
```suggestion
                Collections.unmodifiableCollection(TokenUtil.tokenClaimsToAuthorities(principal.getAttributes(), oktaOAuth2Properties.getGroupsClaim()));
```
Fixes warning

bdemers:
Are there changes to support WebFlux in this PR?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: done

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2ResourceServerAutoConfig.java
arvindkrishnakumar-okta: updated

arvindkrishnakumar-okta:
> Are there changes to support WebFlux in this PR?

It'd be ideal to add it to this PR. Just updated the PR with it.

arvindkrishnakumar-okta:
Ready for review!

oauth2/src/main/java/com/okta/spring/boot/oauth/WebClientUtil.java
```diff
@@ -23,7 +23,7 @@
 
 final class WebClientUtil {
 
-    private static final String USER_AGENT_VALUE = ApplicationInfo.get().entrySet().stream()
+    static final String USER_AGENT_VALUE = ApplicationInfo.get().entrySet().stream()
```
bdemers, line 26: nit: looks like this can still be private?
```suggestion
    private static final String USER_AGENT_VALUE = ApplicationInfo.get().entrySet().stream()
```

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
bdemers: update the test property`okta.oauth2.issuer` to be something that the previous code was expected (so the test functions the same way), then add a new test (if needed for the newer code)

This will ensure the previous functionality continues to work unchanged

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest.groovy
```diff
@@ -69,6 +75,31 @@ class OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest {
         assertThat environment.getProperty(TOKEN_URI), nullValue()
         assertThat environment.getProperty(USER_INFO_URI), nullValue()
         assertThat environment.getProperty(PROVIDER_KEYS_URI), nullValue()
+        assertThat environment.getProperty(RS_CLIENT_ID), nullValue()
+        assertThat environment.getProperty(RS_CLIENT_SECRET), nullValue()
+        assertThat environment.getProperty(RS_INTROSPECTION_URI), nullValue()
+    }
+
+    @Test(expectedExceptions = IllegalArgumentException.class)
```
bdemers, line 83: nit: don't use this pattern of expecting exceptions, unless the test has a single line.
In this case line 91 could throw the `IllegalArgumentException` and the test would still pass

Either use the old try/catch/fail, or the `expect IllegalArgumentException, { yourCodeHere() }` model, similar to what is used in the SDK and other projects: https://github.com/okta/okta-sdk-java/blob/master/impl/src/test/groovy/com/okta/sdk/impl/Util.groovy#L22

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest.groovy
```diff
@@ -69,6 +75,31 @@ class OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest {
...
+        assertThat environment.getProperty(SCOPE, Set), is(["one", "two", "three"] as Set)
+        assertThat environment.getProperty(ISSUER), is("https://issuer.example.com/foobar")
+        assertThat environment.getProperty(RS_KEYS_URI),is("https://issuer.example.com/foobar/v1/keys")
+        assertThat environment.getProperty(AUTHZ_URI), is("https://issuer.example.com/foobar/v1/authorize")
+        assertThat environment.getProperty(TOKEN_URI), is("https://issuer.example.com/foobar/v1/token")
+        assertThat environment.getProperty(USER_INFO_URI), is("https://issuer.example.com/foobar/v1/userinfo")
+        assertThat environment.getProperty(PROVIDER_KEYS_URI), is("https://issuer.example.com/foobar/v1/keys")
+        assertThat environment.getProperty(RS_CLIENT_ID), is("test-client-id")
+        assertThat environment.getProperty(RS_INTROSPECTION_URI), is("https://issuer.example.com/foobar/v1/introspect")
+        assertThat environment.getProperty(RS_CLIENT_SECRET), nullValue()
```
bdemers, line 102: This line should return null instead of throwing an exception

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -92,10 +127,12 @@ private void configureLogin(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Pr
         }
     }
 
-    private void configureResourceServer(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Properties) throws Exception {
+    private void configureResourceServerForOpaqueTokenValidation(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Properties) throws Exception {
 
-        http.oauth2ResourceServer()
-                .jwt().jwtAuthenticationConverter(new OktaJwtAuthenticationConverter(oktaOAuth2Properties.getGroupsClaim()));
+        if (isEmpty(oktaOAuth2Properties.getClientSecret())) {
+            throw new IllegalArgumentException("Missing property 'okta.oauth2.client-secret'");
```
bdemers, line 133: See #207, this should ensure this doesn't happen

bdemers (changes requested):
couple nits,
Some of which should be resolved by #207.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest.groovy
arvindkrishnakumar-okta: noted.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessorTest.groovy
arvindkrishnakumar-okta: noted.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -62,21 +66,52 @@ public void init(HttpSecurity http) throws Exception {
...
+                        } else {
+                            log.debug("JWT configurer is NOT set in OAuth resource server configuration. " +
+                                "Opaque Token validation/introspection will be configured.");
+                            configureResourceServerForOpaqueTokenValidation(http, oktaOAuth2Properties);
+                        }
+                    } else {
+                        String errMsg = "JWT configurer field was not found in OAuth resource server configuration. " +
+                            "Version incompatibility with Spring Security detected." +
+                            "Check https://github.com/okta/okta-spring-boot for project updates.";
+                        throw new RuntimeException(errMsg);
```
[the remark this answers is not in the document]
arvindkrishnakumar-okta: changed to throw `IllegalStateException`.

arvindkrishnakumar-okta:
> couple nits,
> Some of which should be resolved by #207.

fixed.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
bdemers: Also, this is a PKCE test, that requires that no secret be present
(currently, this is a limitation with Spring Security)

bdemers: I'm attempting to push a fix for this

bdemers (approved):
Looks good!

Issue #188, opened 2020-09-16
Add proxy information to README


mraible:
If there's documentation in the README for how to run with a proxy, it'll be easier to answer [questions like this one](https://stackoverflow.com/questions/63913936/proxy-for-spring-security-for-react-oidc).

arvindkrishnakumar-okta:
Fixed.

PR #195, opened 2020-09-23
Update README with Proxy info

arvindkrishnakumar-okta:
### Issue:

https://github.com/okta/okta-spring-boot/issues/188

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ System.setProperty("https.proxyPort", "443");
 
 See [here](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) for the complete list of properties.
 
-Note: This way of setting the Proxy would not work for `webflux` since system properties are presently not supported by Spring's `WebClient` (See [here](https://github.com/spring-projects/spring-security/issues/8882)) for details on this.
+**Note:**  Spring WebFlux (and `WebClient`) do not support these properties. (See spring-projects/spring-security#8882).
 
 # Inject the Okta Java SDK
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ compile 'com.okta.spring:okta-spring-boot-starter'
 
 ## Supporting client side applications - OAuth Implicit flow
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +238,26 @@ You'll be redirected automatically to an Okta login page. Once you successfully
...
+```
+
+or, you could set it programmatically like:
+
+```java
+System.setProperty("https.proxyHost", "https://example-proxy.com");
+System.setProperty("https.proxyPort", "443");
+```
+
+Note: The above samples apply only for `HTTPS` proxy type. Replace `https` with your proxy type as applicable. See [here](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) for the complete list of properties.
```
bdemers, line 259: nit: I don't think we need to mention non-https proxy configuration.  All communication with Okta is over https.

We do need to mention that this does NOT work with WebFlux

bdemers (approved):
Add note about this not working with WebFlux/reactive applications:
https://github.com/spring-projects/spring-security/issues/8882

README.md
arvindkrishnakumar-okta: Addressed.

README.md
```diff
@@ -238,6 +238,28 @@ You'll be redirected automatically to an Okta login page. Once you successfully
...
+or, you could set it programmatically like:
+
+```java
+System.setProperty("https.proxyHost", "https://example-proxy.com");
+System.setProperty("https.proxyPort", "443");
+```
+
+See [here](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) for the complete list of properties.
+
+Note: This way of setting the Proxy would not work for `webflux` since system properties are presently not supported by Spring's `WebClient` (See [here](https://github.com/spring-projects/spring-security/issues/8882)) for details on this.
```
bdemers, line 261:
```suggestion
**Note:**  Spring WebFlux (and `WebClient`) do not support these properties. (See spring-projects/spring-security#8882).
```

PR #212, opened 2020-11-05
Opaque token support for Reactive/WebFlux applications

sergiishamrai-okta:
Opaque token support for Reactive/WebFlux applications


oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2ResourceServerHttpServerAutoConfig.java
```diff
@@ -55,9 +63,25 @@ BeanPostProcessor oktaOAuth2ResourceServerBeanPostProcessor(OktaOAuth2Properties
...
-                http.oauth2ResourceServer().jwt()
+
+                try {
+                    if (TokenUtil.isRootOrgIssuer(oktaOAuth2Properties.getIssuer())) {
+                        log.debug("Opaque Token validation/introspection will be configured.");
+                        http.oauth2ResourceServer().opaqueToken();
+                        return http;
+                    }
+                } catch (MalformedURLException ex) {
+                    log.error(ex.getMessage(), ex);
```
bdemers, line 74: fail here, throw an Illegal Arg or some other type of RuntimeException.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/AutoConfigConditionalTest.groovy
```diff
@@ -501,8 +501,7 @@ class AutoConfigConditionalTest implements HttpMock {
         reactiveContextRunner().withPropertyValues(
             "okta.oauth2.issuer=https://test.example.com/oauth2/custom-as",
             "spring.security.oauth2.client.provider.okta.issuerUri=${mockBaseUrl()}oauth2/custom-as", // work around to not validate the https url
-            "okta.oauth2.client-id=test-client-id",
-            "okta.oauth2.client-secret=test-client-secret")
+            "okta.oauth2.client-id=test-client-id")
```
bdemers, line 504: The existing tests _should_ function the same way to ensure backward compatibility with the previous version
NOTE: It is possible the tests make improper assumptions though, and will need to be updated

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2ResourceServerHttpServerAutoConfig.java
```diff
@@ -55,9 +63,25 @@ BeanPostProcessor oktaOAuth2ResourceServerBeanPostProcessor(OktaOAuth2Properties
         public Object postProcessAfterInitialization(Object bean, String beanName) {
             if (bean instanceof ServerHttpSecurity) {
                 final ServerHttpSecurity http = (ServerHttpSecurity) bean;
-                http.oauth2ResourceServer().jwt()
+
+                try {
+                    if (TokenUtil.isRootOrgIssuer(oktaOAuth2Properties.getIssuer())) {
+                        log.debug("Opaque Token validation/introspection will be configured.");
+                        http.oauth2ResourceServer().opaqueToken();
```
bdemers, line 70: Some type of converter must be used to similar to line 82 below

bdemers (changes requested):
More tests needed,
We need tests that cover these new code branches

This code makes the assumption that setting a client Id and secret should trigger opaque config. This isn't correct.

The user's app _should_ determine this, either by calling `.jwt()` or `.opaqueToken()`.
If one of these methods have NOT been called by the user's app, then it should fallback to the following:

if `isRootOrgIssuer == true` -> opaque
else configure JWT

**NOTE:** opqaueToken configuration should ONLY be used in cases of `rootOrgIssuer` or the user's app explicitly calls `.opaqueToken()`

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2ResourceServerHttpServerAutoConfig.java
```diff
@@ -55,9 +63,25 @@ BeanPostProcessor oktaOAuth2ResourceServerBeanPostProcessor(OktaOAuth2Properties
...
+                        return http;
+                    }
+                } catch (MalformedURLException ex) {
+                    throw new IllegalArgumentException(ex.getMessage());
+                }
+
+                if (!isEmpty(oktaOAuth2Properties.getClientId()) && !isEmpty(oktaOAuth2Properties.getClientSecret())) {
```
arvindkrishnakumar-okta, line 77: This condition could mean we could do JWT as well. I don't think this is a qualifying/sufficient condition to start a Opaque token flow.l

arvindkrishnakumar-okta (changes requested):
See my comment above. Also, need more tests to cover the new code paths.

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2ResourceServerHttpServerAutoConfig.java
arvindkrishnakumar-okta: @sergiishamrai-okta check this out: https://github.com/okta/okta-spring-boot/pull/213/files

mraible:
@sergiishamrai-okta Please rebase to fix conflicting files.

PR #214, opened 2020-11-13
Update to Spring Boot 2.4.0

bdemers:
A few things change in Spring Security 5.4 (used by Spring Boot 2.4)

1. The default WebSecurityAdapter (or equivalent, now configures a "login" or a "resource server", not both by default
  - this change restores that functionality
2. PKCE for unauthenticated clients looks to be NOT configured by default 
  - though I might be missing something there, open question to the Spring Security team on Gitter
3. Default OAuth scopes are NO longer set by Spring Security
  - this change restores that functionality, setting them to the previous default of: `openid, profile, email`
4. It's now easier to set the JWT converter, you can just inject it!
5. Minor test changes (we have tests that poke at the internals of spring sec, these often need to change when implementation details change)

### DIFF
--- a/integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
+++ b/integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
@@ -67,4 +67,5 @@ scenarios:
     - --okta.oauth2.scopes=offline_access
 #    - --okta.oauth2.redirectUri=/authorization-code/callback # not implemented until Spring Sec 5.2
     - --server.servlet.session.tracking-modes=cookie
+    - --spring.security.oauth2.client.registration.okta.client-authentication-method=none
     protectedPath: /oauth2/authorization/okta
--- a/integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
+++ b/integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
@@ -52,7 +52,6 @@ scenarios:
     - --okta.oauth2.issuer=https://localhost:${mockHttpsPort}/oauth2/default
     - --okta.oauth2.clientId=OOICU812
     - --okta.oauth2.clientSecret=VERY_SECRET
-    - --okta.oauth2.scopes=profile,email,openid
 #    - --okta.oauth2.redirectUri=/authorization-code/callback # not implemented until Spring Sec 5.2
     - --server.servlet.session.tracking-modes=cookie
     - --okta.oauth2.postLogoutRedirectUri=/you-are-logged-out
@@ -67,5 +66,4 @@ scenarios:
     - --okta.oauth2.scopes=offline_access
 #    - --okta.oauth2.redirectUri=/authorization-code/callback # not implemented until Spring Sec 5.2
     - --server.servlet.session.tracking-modes=cookie
-    - --spring.security.oauth2.client.registration.okta.client-authentication-method=none
     protectedPath: /oauth2/authorization/okta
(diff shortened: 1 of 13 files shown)


bdemers:
@arvindkrishnakumar-okta can you restart that failing CI job, it looks like a network error on Travis (while downloading maven artifacts)

integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
```diff
@@ -67,4 +67,5 @@ scenarios:
     - --okta.oauth2.scopes=offline_access
 #    - --okta.oauth2.redirectUri=/authorization-code/callback # not implemented until Spring Sec 5.2
     - --server.servlet.session.tracking-modes=cookie
+    - --spring.security.oauth2.client.registration.okta.client-authentication-method=none
```
bdemers, line 70: @mraible looks like this is one of the things that changed in the latest version of Spring Security (I've got a question on the Spring Sec Gitter to confirm this)

This prop needs to be set in order for PKCE (for clients without secrets) to work.  Assuming this is correct, any thoughts on pushing this down to the user like this? or should we re-map it to `okta.oauth2.client-authentication-method=none`, or... should we just detect this condition and configure it automatically (to insure the previous functionality)

(PKCE still isn't used for clients with secrets, that that's another thread...)

arvindkrishnakumar-okta:
> @arvindkrishnakumar-okta can you restart that failing CI job, it looks like a network error on Travis (while downloading maven artifacts)

done

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
```diff
@@ -65,4 +72,26 @@ OidcClientInitiatedLogoutSuccessHandler oidcLogoutSuccessHandler(OktaOAuth2Prope
...
+    @ConditionalOnDefaultWebSecurity
+    static class OAuth2SecurityFilterChainConfiguration {
+        @Bean
+        SecurityFilterChain oauth2SecurityFilterChain(HttpSecurity http) throws Exception {
+            // as of Spring Security 5.4 the default chain uses oauth2Login OR a JWT resource server (NOT both)
+            // this does the same as both defaults merged together (and provides the previous behavior)
+            http.authorizeRequests((requests) -> requests.anyRequest().authenticated());
+            http.oauth2Login(Customizer.withDefaults());
+            http.oauth2Client();
+            http.oauth2ResourceServer(OAuth2ResourceServerConfigurer::jwt);
```
arvindkrishnakumar-okta, line 93: Nice!

mraible: In [this example](https://github.com/spring-projects/spring-security/blob/master/samples/boot/oauth2webclient-webflux/src/main/java/sample/config/SecurityConfig.java), I noticed they use `withDefaults()` for login, client, and resource server. Not sure what that does, but thought I'd mention it.

integration-tests/oauth2-reactive/src/test/resources/testRunner.yml
mraible: I would configure it automatically, then document it. That way, it's a seamless upgrade for folks. As far as I know, you can configure Spring Boot as a resource server with only an issuer. I'm not sure folks will know that an issuer and a client ID means it'll use PKCE. A lot of our JS SDKs have a `pkce` flag that you can turn off. Does it make sense to add an `okta.oauth2.pkce` property?

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
bdemers: I just checked `withDefaults()`, just adds a no-op "customizer"

arvindkrishnakumar-okta (approved):

PR #224, opened 2020-12-04
Improve exception message thrown in startup failure

VitaliiTytarenko-okta:
### Issue:

https://github.com/okta/samples-java-spring/issues/77

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
@@ -35,7 +35,8 @@ public class OktaEnvironmentPostProcessorApplicationListener implements SmartApp
             ConfigurableEnvironment environment = ((ApplicationPreparedEvent) event).getApplicationContext().getEnvironment();
             ValidationResponse validationResponse = ConfigurationValidator.validateIssuer(environment.getProperty("okta.oauth2.issuer"));
             if (!validationResponse.isValid()) {
-                log.warn(validationResponse.getMessage());
+                log.warn(validationResponse.getMessage() + System.lineSeparator() +
+                    "To fix this add the `okta.oauth2.issuer` property to your application environments.");
             }
         }
     }
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
@@ -1,0 +1,36 @@
+package com.okta.spring.boot.oauth.env;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.event.ApplicationPreparedEvent;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.event.SmartApplicationListener;
+import org.springframework.core.Ordered;
+import org.springframework.core.env.ConfigurableEnvironment;
+
+public class OktaEnvironmentPostProcessorApplicationListener implements SmartApplicationListener, Ordered {
+
+    private static final Logger log = LoggerFactory.getLogger(OktaEnvironmentPostProcessorApplicationListener.class);
+    public static final String OKTA_OAUTH2_ISSUER = "okta.oauth2.issuer";
+
+    @Override
+    public void onApplicationEvent(ApplicationEvent event) {
+        if (event instanceof ApplicationPreparedEvent) {
+            ConfigurableEnvironment environment = ((ApplicationPreparedEvent) event).getApplicationContext().getEnvironment();
+            String oktaOauth2Issuer = environment.getProperty(OKTA_OAUTH2_ISSUER);
+            if (oktaOauth2Issuer == null || oktaOauth2Issuer.isEmpty()) {
+                log.warn("Mandatory property `" + OKTA_OAUTH2_ISSUER + "` is missing.");
+            }
+        }
+    }
+
+    @Override
+    public boolean supportsEventType(Class<? extends ApplicationEvent> eventType) {
+        return ApplicationPreparedEvent.class.isAssignableFrom(eventType);
+    }
+
+    @Override
+    public int getOrder() {
+        return Ordered.LOWEST_PRECEDENCE;
+    }
+}
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
@@ -1,5 +1,7 @@
 package com.okta.spring.boot.oauth.env;
 
+import com.okta.commons.configcheck.ConfigurationValidator;
+import com.okta.commons.configcheck.ValidationResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.event.ApplicationPreparedEvent;
@@ -11,15 +13,14 @@ import org.springframework.core.env.ConfigurableEnvironment;
 public class OktaEnvironmentPostProcessorApplicationListener implements SmartApplicationListener, Ordered {
 
     private static final Logger log = LoggerFactory.getLogger(OktaEnvironmentPostProcessorApplicationListener.class);
-    public static final String OKTA_OAUTH2_ISSUER = "okta.oauth2.issuer";
 
     @Override
     public void onApplicationEvent(ApplicationEvent event) {
         if (event instanceof ApplicationPreparedEvent) {
             ConfigurableEnvironment environment = ((ApplicationPreparedEvent) event).getApplicationContext().getEnvironment();
-            String oktaOauth2Issuer = environment.getProperty(OKTA_OAUTH2_ISSUER);
-            if (oktaOauth2Issuer == null || oktaOauth2Issuer.isEmpty()) {
-                log.warn("Mandatory property `" + OKTA_OAUTH2_ISSUER + "` is missing.");
+            ValidationResponse validationResponse = ConfigurationValidator.validateIssuer(environment.getProperty("okta.oauth2.issuer"));
+            if (!validationResponse.isValid()) {
+                log.warn(validationResponse.getMessage());
             }
         }
     }
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017 Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.okta.spring.boot.oauth.env;
 
 import com.okta.commons.configcheck.ConfigurationValidator;
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Okta, Inc.
+ * Copyright 2020-Present Okta, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
@@ -96,7 +96,7 @@ import java.util.Map;
  */
 final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(OktaOAuth2PropertiesMappingEnvironmentPostProcessor.class);
+    private static final Logger log = LoggerFactory.getLogger(OktaOAuth2PropertiesMappingEnvironmentPostProcessor.class);
     private static final String OKTA_OAUTH_PREFIX = "okta.oauth2.";
     private static final String OKTA_OAUTH_ISSUER = OKTA_OAUTH_PREFIX + "issuer";
     private static final String OKTA_OAUTH_CLIENT_ID = OKTA_OAUTH_PREFIX + "client-id";
@@ -204,7 +204,7 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
         @Override
         public Object getProperty(String name) {
             if (OKTA_ISSUER_URI.equals(name) && OKTA_STATIC_DISCOVERY.equals(this.getName()) && !containsProperty(name)) {
-                LOGGER.warn("Mandatory property `" + OKTA_OAUTH_ISSUER + "` is missing");
+                log.warn("Mandatory property `" + OKTA_OAUTH_ISSUER + "` is missing");
             }
 
             return containsProperty(name)
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
@@ -99,6 +99,8 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
     private static final String OKTA_OAUTH_CLIENT_ID = OKTA_OAUTH_PREFIX + "client-id";
     private static final String OKTA_OAUTH_CLIENT_SECRET = OKTA_OAUTH_PREFIX + "client-secret";
     private static final String OKTA_OAUTH_SCOPES = OKTA_OAUTH_PREFIX + "scopes"; // array vs string
+    private static final String OKTA_STATIC_DISCOVERY = "okta-static-discovery";
+    private static final String OKTA_ISSUER_URI = "spring.security.oauth2.client.provider.okta.issuer-uri";
 
     @Override
     public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
@@ -165,9 +167,9 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
         properties.put("spring.security.oauth2.client.provider.okta.token-uri", "${okta.oauth2.issuer}/v1/token");
         properties.put("spring.security.oauth2.client.provider.okta.user-info-uri", "${okta.oauth2.issuer}/v1/userinfo");
         properties.put("spring.security.oauth2.client.provider.okta.jwk-set-uri", "${okta.oauth2.issuer}/v1/keys");
-        properties.put("spring.security.oauth2.client.provider.okta.issuer-uri", "${okta.oauth2.issuer}"); // required for OIDC logout
+        properties.put(OKTA_ISSUER_URI, "${okta.oauth2.issuer}"); // required for OIDC logout
 
-        return new ConditionalMapPropertySource("okta-static-discovery", properties, environment, OKTA_OAUTH_ISSUER);
+        return new ConditionalMapPropertySource(OKTA_STATIC_DISCOVERY, properties, environment, OKTA_OAUTH_ISSUER);
     }
 
     private PropertySource oktaOpaqueTokenPropertySource(Environment environment) {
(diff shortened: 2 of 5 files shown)


sdk/src/main/java/com/okta/spring/boot/sdk/env/OktaSdkPropertiesEnvironmentPostProcessor.java
```diff
@@ -35,11 +35,22 @@
  */
 final class OktaSdkPropertiesEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
+    private static final String OKTA_OAUTH_ISSUER = "okta.oauth2.issuer";
```
bdemers, line 38: This is not correct, `issuer` is NOT required for the Management SDK usage

bdemers (changes requested):
Can you add a description of the issue you are solving? It's possible this logic is just in the wrong location.

The property referenced here is for the OAuth module, but it's in the management SDK related code

VitaliiTytarenko-okta:
@bdemers 
Thanks for your reply.
This PR is for https://oktainc.atlassian.net/browse/OKTA-339695, initial issue is in the description.
`OktaSdkPropertiesEnvironmentPostProcessor` has the lowest priority and runs last, trying to set variables from the `.okta/okta.yml` file (after the postProcessors from the OAuth module have worked). At the end of this EnvironmentPostProcessor I added my check. Otherwise, Spring components will continue their work and raise an exception if the property is empty.
I think it's the right place for it.

bdemers:
@VitaliiTytarenko-okta  The SDK module isn't the right place for this. Forcing the SDK library to use the `issuer` property isn't valid, as that module does NOT use this property.
Spring Boot has a mechanism to set required fields. However, this field SHOULD NOT be required, as it wouldn't be available in some scenarios.  Just like the Spring Security Starter, if the property IS set then the Okta Starter MUST configure itself.

I think there is some miscommunication/misunderstanding here as to how this should work. Message me, and I can walk you through the related code.

arvindkrishnakumar-okta:
> @VitaliiTytarenko-okta The SDK module isn't the right place for this. Forcing the SDK library to use the `issuer` property isn't valid, as that module does NOT use this property.
> Spring Boot has a mechanism to set required fields. However, this field SHOULD NOT be required, as it wouldn't be available in some scenarios. Just like the Spring Security Starter, if the property IS set then the Okta Starter MUST configure itself.
> 
> I think there is some miscommunication/misunderstanding here as to how this should work. Message me, and I can walk you through the related code.

Thanks @bdemers! @VitaliiTytarenko-okta feel free to reach out to me tomorrow and I can assist with this as well.

arvindkrishnakumar-okta (changes requested):
@VitaliiTytarenko-okta You should be adding validation for `okta.oauth2.issuer` property in `oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java` instead. That is where `okta.oauth2` namespace properties are getting transformed into Spring properties.

bdemers:
More clarification:

Validating properties in an EnvironmentPostProcessor is NOT the correct location for this. Configuration properties can be validated via a bean: https://reflectoring.io/validate-spring-boot-configuration-parameters-at-startup/

Requiring `okta.oauth2.issuer` is also not the correct property to use, as these are just aliased to the official Spring Security properties. And customers could be using those instead of the `okta.oauth2` variant.

Failing if the property is NOT set isn't ideal either, Spring Boot starters typically only configure themselves when the related configuration props are set. 

If we just want to set a better error message for a user to see why the OAuth bits didn't load we could add logic to the `ConditionalOn*` classes
https://github.com/okta/okta-spring-boot/tree/master/oauth2/src/main/java/com/okta/spring/boot/oauth

We need to make sure folks have an easy time testing when this lib is on the classpath.

Also, the actual issuer validation needs to happen through the `ConfigurationValidator` class:

https://github.com/okta/okta-commons-java/blob/master/config-check/src/main/java/com/okta/commons/configcheck/ConfigurationValidator.java#L158-L178

IIRC, the first method throws a useful exception, the second returns a response that you can handle (i.e. just add a log warning)


That said, if I'm missing some background on this please let me know.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -175,6 +177,9 @@ private ConditionalMapPropertySource(String name, Map<String, Object> source, En
 
         @Override
         public Object getProperty(String name) {
+            if (OKTA_ISSUER_URI.equals(name) && OKTA_STATIC_DISCOVERY.equals(this.getName()) && !containsProperty(name)) {
+                throw new IllegalArgumentException("Mandatory property `" + OKTA_OAUTH_ISSUER + "` is missing");
```
bdemers, line 181: See my previous comment

arvindkrishnakumar-okta:
@VitaliiTytarenko-okta Per @bdemers's recommendation, let us NOT prevent the server startup if `okta.oauth2.issuer` is NOT set. Instead, lets add validation logic inside the `validate()` of `OktaOAuth2Properties` bean. If the `issuer` is null, log a warning message. Hope we're all on the same page here.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
VitaliiTytarenko-okta: @bdemers 
Unfortunately approach with validators does't work for this property.
We can validate this property: we can define some pattern describing issuer URL, for example. And this case will work perfectly, but only if the issuer property is present.
If issuer-url is absent, then the same exception will be thrown:
`org.springframework.security.oauth2.client.registration.ClientRegistration$Builder.validateAuthorizationCodeGrantType -> java.lang.IllegalArgumentException: tokenUri cannot be empty`
This validation comes before all others.
I’ll continue looking for a better solution...

bdemers: I'm not sure I understand.
Taking a step back, if you run a similar example without the Okta Spring Boot starter (with only Spring Security OAuth), with a missing issuer what happens?

(IIRC Spring Security will fallback to protecting your application with Basic Auth and a generated password, but the application start up will NOT fail)

VitaliiTytarenko-okta: Properties as "token-uri" and "authorization-uri" are mandatory for OAuth 2.0. Application will not start without them, it needs to get tokens from somewhere.
In our Okta App these properties depends on "**issuer-uri**"
https://github.com/okta/okta-spring-boot/blob/8b8a58ca5048abcc9f3b372583a4af4ac628b018/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java#L152
That's why without "issuer-uri" our app fails during configuration phase here:
https://github.com/spring-projects/spring-security/blob/48ef27b80ac449c3221d1da057ab2d803482e903/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java#L673

arvindkrishnakumar-okta: @bdemers I guess @VitaliiTytarenko-okta 's point here is that the standard Springboot Oauth requires `issuer` to be set to construct `token-uri` and `authorization-uri` without which the server would not start. So, for Oktaspringboot we should probably do a similar thing of failing to start the server when `okta.oauth2.issuer` is not set.

bdemers: I just created a new Spring Security project from start.spring.io, I added the two non-okta & non-cloud oauth options. added a simple rest controller, and started the application with no properties set

Spring Security falls back to Basic auth and generates a password:

```txt
Using generated security password: <password here>
```

Let chat about this, I feel like there is a fair bit of miss-communication (I want to make sure it's not me causing this 😉 )

VitaliiTytarenko-okta: This is a pretty simple application without a husk, it has only `org.springframework.boot` dependencies and this is very close to what we have: https://github.com/oktadeveloper/okta-spring-boot-authz-server-example/tree/master/SpringBootOAuthClient
It fails without "token-uri" or "authorization-uri".
Full article: https://developer.okta.com/blog/2019/03/12/oauth2-spring-security-guide

bdemers: That application is expected to fail, due to this line: https://github.com/oktadeveloper/okta-spring-boot-authz-server-example/blob/master/SpringBootOAuthClient/src/main/java/com/okta/spring/SpringBootOAuthClient/SecurityConfiguration.java#L15

Which explicitly tries to configure OAuth.

The user could also programmatically configure the related OAuth bits (without properties), we need to make sure our integration does not break things.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -82,11 +84,14 @@
  */
 final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(OktaOAuth2PropertiesMappingEnvironmentPostProcessor.class);
```
arvindkrishnakumar-okta, line 87:
```suggestion
    private static final Logger log = LoggerFactory.getLogger(OktaOAuth2PropertiesMappingEnvironmentPostProcessor.class);
    
```

to be consistent.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -175,6 +180,9 @@ private ConditionalMapPropertySource(String name, Map<String, Object> source, En
 
         @Override
         public Object getProperty(String name) {
+            if (OKTA_ISSUER_URI.equals(name) && OKTA_STATIC_DISCOVERY.equals(this.getName()) && !containsProperty(name)) {
```
arvindkrishnakumar-okta, line 183: Ideally, this validation logic should be live here: 

https://github.com/okta/okta-spring-boot/blob/8b8a58ca5048abcc9f3b372583a4af4ac628b018/oauth2/src/main/java/com/okta/spring/boot/oauth/config/OktaOAuth2Properties.java#L163

VitaliiTytarenko-okta: The application will crash much earlier, long before this validation.
That's what I'm trying to say: "approach with validators does't work for this property"

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -181,7 +181,7 @@ private ConditionalMapPropertySource(String name, Map<String, Object> source, En
         @Override
         public Object getProperty(String name) {
             if (OKTA_ISSUER_URI.equals(name) && OKTA_STATIC_DISCOVERY.equals(this.getName()) && !containsProperty(name)) {
-                LOGGER.warn("Mandatory property `" + OKTA_OAUTH_ISSUER + "` is missing");
+                log.warn("Mandatory property `" + OKTA_OAUTH_ISSUER + "` is missing");
```
bdemers, line 184: I mentioned this earlier in the thread, but use something like this: https://github.com/okta/okta-commons-java/blob/master/config-check/src/main/java/com/okta/commons/configcheck/ConfigurationValidator.java#L176

The warning should tell the user what they need to do to fix the problem

VitaliiTytarenko-okta: `validateIssuer` method is already used in the OktaOAuth2Properties class. I don't see the need to perform the same check twice, if you mean to use it here.
But when the property is missing - application will fall much earlier, long before validations from OktaOAuth2Properties class.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final Logger log = LoggerFactory.getLogger(OktaEnvironmentPostProcessorApplicationListener.class);
+    public static final String OKTA_OAUTH2_ISSUER = "okta.oauth2.issuer";
+
+    @Override
+    public void onApplicationEvent(ApplicationEvent event) {
+        if (event instanceof ApplicationPreparedEvent) {
+            ConfigurableEnvironment environment = ((ApplicationPreparedEvent) event).getApplicationContext().getEnvironment();
+            String oktaOauth2Issuer = environment.getProperty(OKTA_OAUTH2_ISSUER);
+            if (oktaOauth2Issuer == null || oktaOauth2Issuer.isEmpty()) {
+                log.warn("Mandatory property `" + OKTA_OAUTH2_ISSUER + "` is missing.");
```
bdemers, line 22: Same comment here, improve the log message (see above)

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2017 Okta, Inc.
```
bdemers, line 2:
```suggestion
 * Copyright 2020-Present Okta, Inc.
```

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListener.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(OktaEnvironmentPostProcessorApplicationListener.class);
+
+    @Override
+    public void onApplicationEvent(ApplicationEvent event) {
+        if (event instanceof ApplicationPreparedEvent) {
+            ConfigurableEnvironment environment = ((ApplicationPreparedEvent) event).getApplicationContext().getEnvironment();
+            ValidationResponse validationResponse = ConfigurationValidator.validateIssuer(environment.getProperty("okta.oauth2.issuer"));
+            if (!validationResponse.isValid()) {
+                log.warn(validationResponse.getMessage());
```
bdemers, line 38: You may also need to instruct the user how to apply this setting, as `getMessage` will not mention `okta.oauth2.issuer`

bdemers (changes requested):
Looks pretty good, couple nits

Please add tests before merging

VitaliiTytarenko-okta:
> Looks pretty good, couple nits
> 
> Please add tests before merging

Done!
@bdemers take a look again, please

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+        when(event.getApplicationContext()).thenReturn(context)
+        when(context.getEnvironment()).thenReturn(environment)
+
+        logsAppender.start()
+        ((Logger) LoggerFactory.getLogger(LOGGER_NAME)).addAppender(logsAppender)
+    }
+
+    @AfterMethod
+    void cleanUp() {
+        environment = new MockEnvironment()
```
bdemers, line 60: I'm guessing this should be in an `@BeforeMethod` ?

That said TestNG (and IIRC JUnit 5)  can run tests method in parallel. This can lead to flaky tests, as one test's output could affect the other.

There are probably a few tricks you could you to avoid this, like using a thread-local in your log appended, and making sure the current thread matches.

Though I'm guessing there is a more straightforward solution as well.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+        logsAppender.reset()
+        logsAppender.stop()
```
bdemers, line 63: You should probably remove the appender after the tests too

bdemers (changes requested):
Looks good overall, 
See above comments about threading / flaky test issues.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
VitaliiTytarenko-okta: Agree. Refactoring is done.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,110 @@
...
+import org.testng.annotations.Test
+
+import static org.hamcrest.MatcherAssert.assertThat
+import static org.hamcrest.Matchers.is
+import static org.mockito.Mockito.mock
+import static org.mockito.Mockito.when
+
+class OktaEnvironmentPostProcessorApplicationListenerTest {
+
+    private static final String LOGGER_NAME = "com.okta.spring.boot.oauth.env.OktaEnvironmentPostProcessorApplicationListener"
```
bdemers, line 38: nit: use class name instead of string, If you need the value as a string, call the appropriate method to get it from the class

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,110 @@
...
+        "Follow these instructions to find it: https://bit.ly/finding-okta-domain" + System.lineSeparator() +
+        "To fix this add the `okta.oauth2.issuer` property to your application environments."
+    private static final Logger log = (Logger) LoggerFactory.getLogger(LOGGER_NAME)
+
+    def event = mock(ApplicationPreparedEvent)
+    def context = mock(ConfigurableApplicationContext)
+
+    @AfterMethod
+    void cleanUp() {
+        log.detachAndStopAllAppenders()
```
bdemers, line 49: does this remove the default appender?

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,110 @@
...
+import static org.mockito.Mockito.mock
+import static org.mockito.Mockito.when
+
+class OktaEnvironmentPostProcessorApplicationListenerTest {
+
+    private static final String LOGGER_NAME = "com.okta.spring.boot.oauth.env.OktaEnvironmentPostProcessorApplicationListener"
+    private static final String MSG = "Your Okta Issuer URL is missing. You can copy your domain from the Okta Developer Console. " +
+        "Follow these instructions to find it: https://bit.ly/finding-okta-domain" + System.lineSeparator() +
+        "To fix this add the `okta.oauth2.issuer` property to your application environments."
+    private static final Logger log = (Logger) LoggerFactory.getLogger(LOGGER_NAME)
```
bdemers, line 42: It still looks like this technique the same threading issues as before.  Two tests could be running at the same time, using the same logger. This may result in one test getting the log messages from the other.

One way around this could be to update your `LogsAppender` to only accept messages from a specific thread.  (though there might be a better way of doing this)

If I'm missing something please let me know.

bdemers (changes requested):

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
VitaliiTytarenko-okta: 🆗

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
VitaliiTytarenko-okta: It only removes all previously added appenders from the logger instance.

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
VitaliiTytarenko-okta: Ok, now I've gotten your point.
Thanks!

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
bdemers: can you remove only the appender that is added in this test? otherwise one test could remove the appended of another test

oauth2/src/test/groovy/com/okta/spring/boot/oauth/env/OktaEnvironmentPostProcessorApplicationListenerTest.groovy
```diff
@@ -0,0 +1,119 @@
...
+        environment.getPropertySources().addFirst(new MapPropertySource("test", properties))
+        when(context.getEnvironment()).thenReturn(environment)
+        when(event.getApplicationContext()).thenReturn(context)
+    }
+
+    private static final class LogsAppender extends ListAppender<ILoggingEvent> {
+        private final String threadName
+
+        LogsAppender() {
+            threadName = Thread.currentThread().getName()
```
bdemers, line 97: nit: The thread name isn't guaranteed to be unique

VitaliiTytarenko-okta: Changed to `thread = Thread.currentThread()`

bdemers (approved):

arvindkrishnakumar-okta (approved):

Issue #280, opened 2021-05-03
Add support for Spring Boot 2.5


mraible:
### I'm submitting a
  - [ ] bug report
  - [x] feature request

### Background info

If I try to create a Spring Boot app on [start.spring.io](https://start.spring.io), it says I can't use Okta.

<img width="1280" alt="Screen Shot 2021-05-03 at 3 07 16 PM" src="https://user-images.githubusercontent.com/17892/116934156-d0af9c00-ac21-11eb-815a-bf279491795f.png">

However, if I download Spring Boot 2.5 RC1 (with a "web" dependency), and modify the `DemoApplication.java` to match the following, everything works as expected.

```java
package com.example.demo;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.security.core.annotation.AuthenticationPrincipal;
import org.springframework.security.oauth2.core.oidc.user.OidcUser;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

@SpringBootApplication
@RestController
public class DemoApplication {

    public static void main(String[] args) {
        SpringApplication.run(DemoApplication.class, args);
    }

    @GetMapping("/")
    public String hello(@AuthenticationPrincipal OidcUser user) {
        return "Hello, " + user.getFullName();
    }
}
```

### Expected behavior

I should be able to create a Spring Boot 2.5 app on start.spring.io and include Okta.

### SDK Version

2.0.1

aarongranick-okta:
Thanks for the report, @mraible. Looks like @bdemers is working on a PR to resolve this.

PR #281, opened 2021-05-03
Testing build with Spring Boot 2.5.1

bdemers:
Fixes: #280

From commits:

Update to Spring Boot 2.5

    With 2.5 a JwtDecoder bean is now initialized when `spring.security.oauth2.resourceserver.jwt.issuer-uri` is set.  There were a few invalid test assumptions that were setting this property (likely a copypasta eror)
    For example the OAuth Code Flow ITs check to make sure requests to the /keys endpoint is NOT called (not needed for non OIDC or jwt resource server usage).

    Another fix with a fragile test which detects the internal configuration of Spring Security.

    OktaOAuth2Configurer had an unused else block, which caused a problem with 2.5 when a resource server was not configured.

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
@@ -254,7 +254,7 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
 
             // only support this key
             if (OKTA_OAUTH_ISSUER_WITH_PATH.equals(key)) {
-                // issuer could be null
+                // issuer could be null (only resolve properties after checking if the key is `OKTA_OAUTH_ISSUER_WITH_PATH`)
                 return Optional.ofNullable(environment.getProperty(OKTA_OAUTH_ISSUER))
                     .map(issuer -> {
                         // Check if URL is a Okta Org Authorization Server
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
@@ -15,6 +15,7 @@
  */
 package com.okta.spring.boot.oauth.env;
 
+import org.jetbrains.annotations.NotNull;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
@@ -97,6 +98,7 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
 
     private static final String OKTA_OAUTH_PREFIX = "okta.oauth2.";
     private static final String OKTA_OAUTH_ISSUER = OKTA_OAUTH_PREFIX + "issuer";
+    private static final String OKTA_OAUTH_ISSUER_WITH_PATH = OKTA_OAUTH_ISSUER + "-with-path";
     private static final String OKTA_OAUTH_CLIENT_ID = OKTA_OAUTH_PREFIX + "client-id";
     private static final String OKTA_OAUTH_CLIENT_SECRET = OKTA_OAUTH_PREFIX + "client-secret";
     private static final String OKTA_OAUTH_SCOPES = OKTA_OAUTH_PREFIX + "scopes"; // array vs string
@@ -110,6 +112,7 @@ final class OktaOAuth2PropertiesMappingEnvironmentPostProcessor implements Envir
         // default scopes, as of Spring Security 5.4 default scopes are no longer added, this restores that functionality
         environment.getPropertySources().addLast(defaultOktaScopesSource(environment));
         // okta's endpoints can be resolved from an issuer
+        environment.getPropertySources().addLast(new OktaIssuerWithPathPropertySource(environment));
         environment.getPropertySources().addLast(oktaStaticDiscoveryPropertySource(environment));
         environment.getPropertySources().addLast(oktaOpaqueTokenPropertySource(environment));
         environment.getPropertySources().addLast(oktaRedirectUriPropertySource(environment));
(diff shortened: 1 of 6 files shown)


bdemers:
ugg, it looks like Travis CI might be blocked from accessing the Spring Maven repo?

mraible:
@bdemers Do you need to add the plugin repo too? If you create a new 2.5 app on start.spring.io, it's got both:

```xml
  <repositories>
    <repository>
      <id>spring-milestones</id>
      <name>Spring Milestones</name>
      <url>https://repo.spring.io/milestone</url>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
    </repository>
  </repositories>
  <pluginRepositories>
    <pluginRepository>
      <id>spring-milestones</id>
      <name>Spring Milestones</name>
      <url>https://repo.spring.io/milestone</url>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
    </pluginRepository>
  </pluginRepositories>
```

bdemers:
Probably, I'll have to try it locally I'm not sure why Travis hit a 401 from that repo 🤷

bdemers:
Doh... i think it was a `http` vs `https` copypasta error

mraible:
Spring Boot 2.5 is out and this is still an issue on start.spring.io.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -158,18 +161,14 @@ private PropertySource oktaRedirectUriPropertySource(Environment environment) {
...
-        properties.put("spring.security.oauth2.client.provider.okta.user-info-uri", "${okta.oauth2.issuer}" + oauth2Suffix + "/v1/userinfo");
-        properties.put("spring.security.oauth2.client.provider.okta.jwk-set-uri", "${okta.oauth2.issuer}" + oauth2Suffix + "/v1/keys");
+        properties.put("spring.security.oauth2.resourceserver.jwt.jwk-set-uri", "${" + OKTA_OAUTH_ISSUER_WITH_PATH + "}/v1/keys");
+        properties.put("spring.security.oauth2.client.provider.okta.authorization-uri", "${" + OKTA_OAUTH_ISSUER_WITH_PATH + "}/v1/authorize");
+        properties.put("spring.security.oauth2.client.provider.okta.token-uri", "${" + OKTA_OAUTH_ISSUER_WITH_PATH + "}/v1/token");
+        properties.put("spring.security.oauth2.client.provider.okta.user-info-uri", "${" + OKTA_OAUTH_ISSUER_WITH_PATH + "}/v1/userinfo");
+        properties.put("spring.security.oauth2.client.provider.okta.jwk-set-uri", "${" + OKTA_OAUTH_ISSUER_WITH_PATH + "}/v1/keys");
         properties.put("spring.security.oauth2.client.provider.okta.issuer-uri", "${okta.oauth2.issuer}"); // required for OIDC logout
```
bdemers, line 172: This should probably be replaced with the default discovery logic in the near future.  We just need to make sure (or force) the non OIDC usage (offline claim) to use discovery.

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -158,18 +161,14 @@ private PropertySource oktaRedirectUriPropertySource(Environment environment) {
     }
 
     private PropertySource oktaStaticDiscoveryPropertySource(Environment environment) {
-        String oauth2Suffix = Optional.ofNullable(environment.getProperty("okta.oauth2.issuer"))
-            .filter(issuer -> issuer.contains("/oauth2")) // Check if URL is a Custom Authorization Server
-            .map(s -> "") // if so, leave URL as is
-            .orElse("/oauth2"); // for the Okta Org Authorization Server, /oauth2 needs to be appended
```
bdemers, line 164: This was mixing eager property initialization with delayed (which may be required for spring cloud config, or similar property resolution)

rfigueroa:
there is a dependency on [Jackson BOM](https://github.com/okta/okta-spring-boot/blob/master/pom.xml#L58-L64), is that still needed? since now that the library is using most recent spring boot

bdemers:
@rfigueroa I'd love to remove it, but given the rate of security-related releases for jackson-databind, I'm hesitant to remove it.  

That said, it's something we should keep an eye on (and remove any unnecessary deps when possible)

rfigueroa:
I think once you start using boot 2.5.1 it would be possible to remove it, as boot 2.5.1 already uses the version of [Jackson BOM](https://github.com/spring-projects/spring-boot/blob/v2.5.1/spring-boot-project/spring-boot-dependencies/build.gradle#L530) you have here.

bdemers:
@rfigueroa possibly, Jackson has been in the Spring Boot BOM for a while. The main issue is that if we need to cut a release of this lib, and the Spring Boot BOM, hasn't been released yet with a newer Jackson version.  This was happening frequently with previous versions of Jackson (2.10 IIRC, and the series of security-related releases for that minor).

mraible (approved):

VitaliiTytarenko-okta (approved):

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
```diff
@@ -234,4 +233,48 @@ public Object getProperty(String name) {
...
+                        if (!issuer.contains("/oauth2")) {
+                            // for the Okta Org Authorization Server, /oauth2 needs to be appended
+                            issuer += "/oauth2";
+                        }
+                        // for all other issuers leave the property as is
+                        return issuer;
+                    })
+                    // otherwise return null
+                    .orElse(null);
+            }
```
VitaliiTytarenko-okta, line 270:
```suggestion
            String issuer = environment.getProperty(OKTA_OAUTH_ISSUER);

            // only support this key
            if (OKTA_OAUTH_ISSUER_WITH_PATH.equals(key) && issuer != null) {
                if (!issuer.contains("/oauth2")) {
                    // for the Okta Org Authorization Server, /oauth2 needs to be appended
                    issuer += "/oauth2";
                }
                // for all other issuers leave the property as is
                return issuer;
            }
```
Feel free to leave it as it is, but in my opinion, this looks better

shuowu-okta (approved):

oauth2/src/main/java/com/okta/spring/boot/oauth/env/OktaOAuth2PropertiesMappingEnvironmentPostProcessor.java
bdemers: @VitaliiTytarenko-okta I agree your block looks better, but after trying it out I realized it didn't work because creates an infinite loop because `environment.getProperty("okta.oauth2.issuer")` may contain the other nested properties.
So, it's only safe to lookup that property after asserting the current key is `OKTA_OAUTH_ISSUER_WITH_PATH`)

I did update the comment though!

PR #292, opened 2021-06-09
Fix cve

arvindkrishnakumar-okta:


src/owasp/owasp-suppression.xml
```diff
@@ -36,4 +36,16 @@
         <cve>CVE-2012-5055</cve>
     </suppress>
 
+    <!-- should be resolved when we upgrade to spring-boot-starter-reactor-netty version 2.5 -->
```
arvindkrishnakumar-okta, line 39: We have more changes in scope when we upgrade to 2.5. Upgrading to 2.5 would resolve the below CVEs though. Since we're not ready to upgrade to 2.5 at this point, I've added these CVEs to suppression list.

bdemers (changes requested):
Don't ignore this one, update to Boot `2.4.6` which contains the fix for this:
https://spring.io/blog/2021/05/20/spring-boot-2-4-6-available-now

The suppression file should only be used for false positives or in rare cases build-time only dependencies like swagger (which does not affect the runtime of downstream dependencies)

arvindkrishnakumar-okta:
Closing this in favor of #293

PR #303, opened 2021-06-17
Removing orphaned/unused modules after splitting up integration tests

mraible:
Related to https://github.com/okta/okta-spring-boot/pull/261.


bdemers (approved):
Good catch!

PR #311, opened 2021-06-22
Change oidcUserService to use Qualifer bean

mraible:
As learned with @joshlong and @bdemers on Twitch today.


oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
```diff
@@ -85,4 +88,4 @@ SecurityFilterChain oauth2SecurityFilterChain(HttpSecurity http) throws Exceptio
             return http.build();
         }
     }
-}
\ No newline at end of file
+}
```
bdemers, line 91: That's what I get for editing through the GH UI...

bdemers (approved):

PR #338, opened 2021-09-07
Add works with OpenJDK badge

mraible:
https://foojay.io/today/works-with-openjdk/


JayNewstrom:
Thanks for the PR. @VitaliiTytarenko-okta or @arvindkrishnakumar-okta could you take a look?

bdemers:
Nice! we should add this to all of our Java repos, IIRC, they all currently test with OpenJDK 8+11 
(I'll open an issue to add 17 into the mix)

VitaliiTytarenko-okta (approved):

PR #340, opened 2021-09-14
Upgrade to springboot 2.5.4

arvindkrishnakumar-okta:


oauth2/pom.xml
```diff
@@ -39,6 +39,12 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
```
arvindkrishnakumar-okta, line 42: Context: https://github.com/okta/okta-spring-boot/pull/331#issuecomment-918758294

mraible: A better import might be the Validation API:

```xml
<dependency>
    <groupId>javax.validation</groupId>
    <artifactId>validation-api</artifactId>
</dependency>
```

Then, change the import to:

```java
import javax.validation.constraints.NotNull;
```

WDYT @bdemers?

I tried this and all tests pass.

mraible:
NOTE: It seems that `OktaJwtAuthenticationConverter#extractAuthorities()` overrides a deprecated method.

oauth2/pom.xml
arvindkrishnakumar-okta: Sounds good to me, updated PR.

bdemers: 💯 The version will be managed automatically too.

mraible (approved):

bdemers (approved):

bdemers:
There were two test deps out of date rest-assured and TestNG,
The rest-assured bump is here: #342 

Updating TestNG will require updating Surefire and Failsafe in the parent pom (to 3.x), but that shouldn't block a release

PR #343, opened 2021-09-14
Address cves

arvindkrishnakumar-okta:


JayNewstrom (approved):

src/owasp/owasp-suppression.xml
```diff
@@ -47,4 +47,19 @@
         <cve>CVE-2020-10693</cve>
     </suppress>
 
+    <!-- false positive, see https://github.com/jeremylong/DependencyCheck/issues/3606
```
bdemers, line 50:
```suggestion
    <!-- false positive, see https://github.com/jeremylong/DependencyCheck/issues/3594
```
nit

bdemers (approved):
nit: add a `TODO` comment around the tomcat deps, so we know to remove them (likely in the next patch release of Spring Boot)

Issue #192, opened 2020-09-21
Okta Spring Boot starter doesn't work with GraalVM


mraible:
Steps to reproduce:

Create a new Spring Boot app using [HTTPie](https://httpie.org):

```
http https://start.spring.io/starter.zip \
     dependencies==web,okta \
     packageName==com.okta.rest \
     name=spring-boot \
     type=maven-project \
     -o spring-boot.zip
```

Then, add a `HelloController`:

```java
package com.okta.rest.controller;

import org.springframework.security.core.annotation.AuthenticationPrincipal;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

import java.security.Principal;

@RestController
public class HelloController {

    @GetMapping("/hello")
    public String hello(@AuthenticationPrincipal Principal principal) {
        return "Hello, " + principal.getName() + "!";
    }
}
```

Configure it to be an OAuth 2.0 resource server by adding an issuer to `application.properties`:

```
okta.oauth2.issuer=https://dev-133337.okta.com/oauth2/default
```

All the following should work at this point (I got an access token from oidcdebugger.com):

```
$ mvn spring-boot:run
$ http :8080/hello
$ TOKEN=eyJraWQiOiJxOE1QMjFNNHZCVmxOSkxGbFFWNlN...
$ http :8080/hello Authorization:"Bearer $TOKEN"
```

To add "build native image support", I followed [these docs](https://repo.spring.io/milestone/org/springframework/experimental/spring-graalvm-native-docs/0.8.0/spring-graalvm-native-docs-0.8.0.zip!/reference/index.html#_configure_the_maven_plugin).

I upgraded my app to use Spring Boot v2.4.0-M2.

```xml
<parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>2.4.0-M2</version>
    <relativePath/> <!-- lookup parent from repository -->
</parent>
```

I updated my configuration to avoid proxies:

```java
@SpringBootApplication(proxyBeanMethods = false)
public class Application {

	public static void main(String[] args) {
		SpringApplication.run(Application.class, args);
	}
}
```

I added the milestone repos to my `pom.xml`:

```xml
<repositories>
    <repository>
        <id>spring-milestones</id>
        <name>Spring Milestones</name>
        <url>https://repo.spring.io/milestone</url>
    </repository>
</repositories>
<pluginRepositories>
[...]

PR #344, opened 2021-09-14
proxyBeanMethods = false not needed since Spring Native 0.8.3

mraible:
See https://github.com/spring-projects-experimental/spring-native/releases/tag/0.8.3


mraible:
This could be relevant, but it is from a while ago. https://github.com/okta/okta-spring-boot/issues/192#issuecomment-745130104

@sdeleuze Is it still a good idea to use `proxyBeanMethods = false` for libraries or Spring Boot starters?

arvindkrishnakumar-okta (approved):

Issue #445, opened 2022-05-22
When threre exist other Environment instance, OktaOAuth2Configurer in okta-springboot-starter will cause NoUniqueBeanDefinitionException


fjdingsd:
### Background info

When I have my own org.springframework.core.env.Environment instantce or exist in other third party dependency,  in the progress of  WebSecurityConfiguration.springSecurityFilterChain bean building ,OktaOAuth2Configurer in okta-springboot-starter initializes itself and to use **context.getBean(Environment.class))** will cause NoUniqueBeanDefinitionException

```java
final class OktaOAuth2Configurer extends AbstractHttpConfigurer<OktaOAuth2Configurer, HttpSecurity> {

    private static final Logger log = LoggerFactory.getLogger(OktaOAuth2Configurer.class);

    @SuppressWarnings("rawtypes")
    @Override
    public void init(HttpSecurity http) throws Exception {

        ApplicationContext context = http.getSharedObject(ApplicationContext.class);

        // make sure OktaOAuth2Properties are available
        if (!context.getBeansOfType(OktaOAuth2Properties.class).isEmpty()) {
            OktaOAuth2Properties oktaOAuth2Properties = context.getBean(OktaOAuth2Properties.class);

            // Auth Code Flow Config

            // if OAuth2ClientProperties bean is not available do NOT configure
            OAuth2ClientProperties.Provider propertiesProvider;
            OAuth2ClientProperties.Registration propertiesRegistration;
            if (!context.getBeansOfType(OAuth2ClientProperties.class).isEmpty()
                && (propertiesProvider = context.getBean(OAuth2ClientProperties.class).getProvider().get("okta")) != null
                && (propertiesRegistration = context.getBean(OAuth2ClientProperties.class).getRegistration().get("okta")) != null
                && !isEmpty(propertiesProvider.getIssuerUri())
                && !isEmpty(propertiesRegistration.getClientId())) {
                // configure Okta user services
                configureLogin(http, oktaOAuth2Properties, context.getBean(Environment.class));
```

and here's my own Environment definition:

```java
@Component("kmsEnvironment")
public class KMSEnvironment extends StandardEnvironment implements ConfigurableEnvironment , EnvironmentAware{
  // ... something else
}
```

and error log (kmsEnvironment is my own Environment definition):
```java
***************************
APPLICATION FAILED TO START
***************************

Description:

[...]

PR #447, opened 2022-05-26
Obvious Fix the NoUniqueBeanDefinitionException when get Environment bean

fjdingsd:
Obvious Fix：
Fix the occurrence of NoUniqueBeanDefinitionException when there are multiple `org.springframework.core.env.Environment` instance bean during `OktaOAuth2Configurer.init()` to call `context.getBean(Class<T> requiredType)` where only class type as needed. More details see #445

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
@@ -62,7 +62,7 @@ final class OktaOAuth2Configurer extends AbstractHttpConfigurer<OktaOAuth2Config
                 && !isEmpty(propertiesProvider.getIssuerUri())
                 && !isEmpty(propertiesRegistration.getClientId())) {
                 // configure Okta user services
-                configureLogin(http, oktaOAuth2Properties, context.getBean(Environment.class));
+                configureLogin(http, oktaOAuth2Properties, context.getEnvironment());
 
                 // check for RP-Initiated logout
                 if (!context.getBeansOfType(OidcClientInitiatedLogoutSuccessHandler.class).isEmpty()) {


bdemers (approved):
Awesome!

arvindkrishnakumar-okta:
@fjdingsd Awesome, thanks!

arvindkrishnakumar-okta:
@fjdingsd Can you please sign the [CLA](https://developer.okta.com/cla/) here?

arvindkrishnakumar-okta (approved):

fjdingsd:
Hi, @arvindkrishnakumar-okta  and bdemers, thanks for approval !

Here, I'm a little confused about the way of OKTA's CLA signature :
1. github username and profile name, which is the Printed Name? (Take you as an example, is arvindkrishnakumar-okta or Arvind Krishnakumar ? )
2. Is allowed to electronically sign the CLA PDF ?

bdemers:
"Printed name" would be your legal name, IANAL, but it should probably be something like "Name, printed" as opposed to "signature", which would be "Name, cursive" (though that's bit of a simplification).

That said, in hindsight, this could probably be marked as an "obvious fix" (per this [guide](https://developer.okta.com/cla/)), and wouldn't need the CLA signed.

fjdingsd:
> That said, in hindsight, this could probably be marked as an "obvious fix" (per this [guide](https://developer.okta.com/cla/)), and wouldn't need the CLA signed.

That means I don't need to sign CLA for this commit, right? If so, Should I add the word 'Obvious Fix' to my first conversation comment ? Or new pull request with  the word 'Obvious Fix' in commit comment ?

Please lead me on, Thanks!

bdemers:
For completenesses you can just update the description. And we should be able to merge it when we are back from the long weekend.

thanks again!!

fjdingsd:
I've update the title and description together. Have a nice weekend!

arvindkrishnakumar-okta:
@fjdingsd thanks, I'm noticing another issue upgrading the SDK to Springboot v2.7.0 that I need to fix first.

PR #491, opened 2022-10-29
Update latest-spring.yml

arvindkrishnakumar-okta:


.github/workflows/latest-spring.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     name: Java ${{ matrix.java }}
     strategy:
       matrix:
-        java: [8, 11]
+        java: [11]
```
mraible, line 27: This starter works with Spring Boot 3, so I think this should be updated to include 17.

mraible (approved):

.github/workflows/latest-spring.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     name: Java ${{ matrix.java }}
     strategy:
       matrix:
-        java: [8, 11]
+        java: [11, 17]
```
bdemers, line 27: The latest v3 version is going to require 17+, my suggestion is to leave the matrix array, as we may need want to add newer versions like 21 in the future


```suggestion
        java: [17]
```

bdemers (approved):

PR #501, opened 2022-11-16
Add support for `acr_values`, `prompt` and `enroll_amr_values` params

arvindkrishnakumar-okta:
Related PR: https://github.com/okta/okta-oidc-tck/pull/173/files


laura-rodriguez (approved):
LGTM

bdemers (changes requested):
This looks like a good draft, but it assumes that any configured "enrollment" values should be sent on every authorization request.
Instead of specialized requests. e.g. a user already has a session and needs to be redirected to enroll/prompt for factor.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2Configurer.java
```diff
@@ -180,7 +182,12 @@ private void configureLogin(HttpSecurity http, OktaOAuth2Properties oktaOAuth2Pr
             http.oauth2Login().redirectionEndpoint().baseUri(redirectUri);
         }
 
-        Okta.configureOAuth2WithPkceAndAuthReqParams(http, clientRegistrationRepository);
+        // check if user is authenticated, if so then prompt for enroll (populate custom auth req parameters for enroll flow)
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+
+        if (authentication != null && authentication.getPrincipal() != null) {
+            Okta.configureOAuth2WithPkceAndAuthReqParams(http, clientRegistrationRepository);
+        }
```
bdemers, line 190: This logic will execute as the application is starting so the current auth would always be `null`.

To execute these new flows ("prompt to enroll in factor" and "prompt for factor"), there needs to be a new high-level entry point (either programmatically or based on a redirect).

Checking if a user is already authenticated before adding these args may be appropriate here (guessing, but I'm not 100% sure).  If a user visits /login/oauth in a spring app with an existing session that behavior _shouldn't_ change, that doesn't indicate that the app is requiring a factor (it could just be how the app's links are structured).

PR #512, opened 2022-12-22
Replace deprecated WebSecurityConfigurerAdapter with SecurityFilterChain usage

bdemers:


examples/siw-jquery/src/main/java/com/okta/spring/example/ImplicitFlowApplication.java
```diff
@@ -17,30 +17,31 @@
...
     @Configuration
-    static class OktaOAuth2WebSecurityConfigurerAdapter extends WebSecurityConfigurerAdapter {
+    static class SecurityConfig {
 
-        @Override
-        protected void configure(HttpSecurity http) throws Exception {
+        @Bean
+        SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+            http.authorizeHttpRequests()
+                .antMatchers("/", "/index.html", "/sign-in-widget-config").permitAll()
```
mraible, line 40: `requestMatchers` is a recommended alternative to `antMatchers`.

bdemers: I ran into a problem with this one.  IIRC, for v3, there is a `.requestMatchers(String...)` but the corresponding method signature isn't in 2.x 🤷 

I have some local changes with the related changes (along with updating the pom to use Spring Boot v3, but I ran into unrelated issues with WireMock (used to create a mock OAuth server).

Once we get the WireMock issues sorted out, it might be possible to add another set of Integration tests, to run against v3 🤔

mraible: I didn't realize you were trying to keep Spring Boot 2.x compatibility. Makes sense now.

arvindkrishnakumar-okta (approved):

Issue #596, opened 2023-08-31
Using this starter with Auth0 results in an opaque access token with `oauth2Login()`


mraible:
### I'm submitting a
  - [ x ] bug report
  - [ ] feature request

### Background info

If I use this starter in a Spring Boot 3.1.3 app, I don't get a JWT as an access token. I believe this is because it doesn't send the audience parameter to the `/authorize` endpoint. With Okta, I do get an access token as a JWT. Here's how I'm testing it:

```java
package com.example.apigateway;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.context.event.ContextRefreshedEvent;
import org.springframework.context.event.EventListener;
import org.springframework.core.env.AbstractEnvironment;
import org.springframework.core.env.EnumerablePropertySource;
import org.springframework.core.env.Environment;
import org.springframework.core.env.MutablePropertySources;
import org.springframework.security.core.annotation.AuthenticationPrincipal;
import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
import org.springframework.security.oauth2.core.OAuth2AccessToken;
import org.springframework.security.oauth2.core.oidc.user.OidcUser;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

import java.security.Principal;
import java.util.Arrays;
import java.util.stream.StreamSupport;

@SpringBootApplication
public class ApiGatewayApplication {

    private static final Logger LOGGER = LoggerFactory.getLogger(ApiGatewayApplication.class);

    public static void main(String[] args) {
        SpringApplication.run(ApiGatewayApplication.class, args);
    }

    @EventListener
    public void handleContextRefresh(ContextRefreshedEvent event) {
        final Environment env = event.getApplicationContext()
                .getEnvironment();

        LOGGER.info("Active profiles: {}", Arrays.toString(env.getActiveProfiles()));

        final MutablePropertySources sources = ((AbstractEnvironment) env).getPropertySources();

        StreamSupport.stream(sources.spliterator(), false)
                .filter(ps -> ps instanceof EnumerablePropertySource)
                .map(ps -> ((EnumerablePropertySource) ps).getPropertyNames())
                .flatMap(Arrays::stream)
                .distinct()
[...]

PR #600, opened 2023-09-11
Add support for audience parameter

mraible:
Fixes https://github.com/okta/okta-spring-boot/issues/596.

### DIFF
--- a/oauth2/src/main/java/com/okta/spring/boot/oauth/Okta.java
+++ b/oauth2/src/main/java/com/okta/spring/boot/oauth/Okta.java
@@ -15,6 +15,7 @@
  */
 package com.okta.spring.boot.oauth;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
@@ -24,8 +25,10 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository;
 import org.springframework.security.oauth2.client.web.DefaultOAuth2AuthorizationRequestResolver;
 import org.springframework.security.oauth2.client.web.server.DefaultServerOAuth2AuthorizationRequestResolver;
+import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizationRequestResolver;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.server.resource.BearerTokenError;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
 import org.springframework.security.oauth2.server.resource.web.server.BearerTokenServerAuthenticationEntryPoint;
@@ -35,6 +38,8 @@ import org.springframework.security.web.server.util.matcher.MediaTypeServerWebEx
 import org.springframework.security.web.util.matcher.MediaTypeRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.web.accept.HeaderContentNegotiationStrategy;
+
+import java.util.function.Consumer;
 
 import static org.springframework.security.oauth2.client.web.OAuth2AuthorizationRequestCustomizers.withPkce;
 
@@ -97,6 +102,12 @@ public final class Okta {
         authorizationRequestResolver.setAuthorizationRequestCustomizer(withPkce());
         // enable oauth2 login that uses PKCE
         http.oauth2Login().authorizationRequestResolver(authorizationRequestResolver);
+        // enable passing the audience parameter
+        http.oauth2Login(oauth2 -> oauth2
+            .authorizationRequestResolver(
+                authorizeRequestResolver(clientRegistrationRepository)
+            )
+        );
 
         return http;
     }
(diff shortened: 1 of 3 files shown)


oauth2/src/main/java/com/okta/spring/boot/oauth/Okta.java
```diff
@@ -135,6 +130,61 @@ public static HttpSecurity configureOAuth2WithPkce(HttpSecurity http, ClientRegi
...
+                reactiveAuthzRequestResolver(clientRegistrationRepository, properties.getAudience())
+            )
+        );
+
+        return http;
+    }
+
+    private static boolean isDefaultAudience(OktaOAuth2Properties properties) {
+        String audience = properties.getAudience();
+        return audience == null || audience.isEmpty() || audience.equals("api://default");
```
mraible, line 159: This means that `api://default` will never be sent if it's the audience. I tried to make it so it only happens when it's Auth0, but was unable to figure out how to do it.

oauth2/src/main/java/com/okta/spring/boot/oauth/OktaOAuth2AutoConfig.java
```diff
@@ -79,11 +79,12 @@ OAuth2UserService<OidcUserRequest, OidcUser> oidcUserService(
     static class OAuth2SecurityFilterChainConfiguration {
 
         @Bean
-        SecurityFilterChain oauth2SecurityFilterChain(HttpSecurity http, ClientRegistrationRepository clientRegistrationRepository) throws Exception {
+        SecurityFilterChain oauth2SecurityFilterChain(HttpSecurity http, ClientRegistrationRepository clientRegistrationRepository, OktaOAuth2Properties oktaOAuth2Properties) throws Exception {
             // as of Spring Security 5.4 the default chain uses oauth2Login OR a JWT resource server (NOT both)
             // this does the same as both defaults merged together (and provides the previous behavior)
             http.authorizeRequests((requests) -> requests.anyRequest().authenticated());
             Okta.configureOAuth2WithPkce(http, clientRegistrationRepository);
+            Okta.configureOAuth2WithAudience(http, clientRegistrationRepository, oktaOAuth2Properties);
```
mraible, line 87: The audience overrides the PKCE parameters added on the previous line. Is there a way to chain these together so both parameters are added?

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2AutoConfig.java
```diff
@@ -69,13 +64,13 @@ OidcReactiveOAuth2UserService oidcUserService(Collection<AuthoritiesProvider> au
     }
 
     @Bean
-    @ConditionalOnBean(ReactiveJwtDecoder.class)
```
mraible, line 72: Removing this was necessary to invoke this configuration when no `SecurityConfiguration` exists.

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2AutoConfig.java
```diff
@@ -69,13 +64,13 @@ OidcReactiveOAuth2UserService oidcUserService(Collection<AuthoritiesProvider> au
...
     @ConditionalOnMissingBean(SecurityWebFilterChain.class)
-    SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http, ReactiveJwtDecoder jwtDecoder, ReactiveClientRegistrationRepository clientRegistrationRepository) {
+    SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http, ReactiveJwtDecoder jwtDecoder, ReactiveClientRegistrationRepository clientRegistrationRepository, OktaOAuth2Properties properties) throws Exception {
         // as of Spring Security 5.4 the default chain uses oauth2Login OR a JWT resource server (NOT both)
         // this does the same as both defaults merged together (and provides the previous behavior)
         http.authorizeExchange().anyExchange().authenticated();
         Okta.configureOAuth2WithPkce(http, clientRegistrationRepository);
+        Okta.configureOAuth2WithAudience(http, clientRegistrationRepository, properties);
```
mraible, line 73: This line clobbers the previous lines added parameters. We need to figure out a way to chain them together.

oauth2/src/main/java/com/okta/spring/boot/oauth/Okta.java
```diff
@@ -162,22 +212,31 @@ static String statusAsString(HttpStatus status) {
...
 
-    @Value("${okta.oauth2.audience:}")
-    private static String audience;
-
-    private static ServerOAuth2AuthorizationRequestResolver authorizeRequestResolver(
-        ReactiveClientRegistrationRepository clientRegistrationRepository) {
+    private static ServerOAuth2AuthorizationRequestResolver reactiveAuthzRequestResolver(
+        ReactiveClientRegistrationRepository clientRegistrationRepository, String audience) {
 
         DefaultServerOAuth2AuthorizationRequestResolver authorizationRequestResolver =
```
mraible, line 218: How do you get the `authorizationRequestResolver` if it already exists? If I could figure that out, I could probably solve the chaining problem where the audience parameters overwrite the PKCE parameters.

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2AutoConfig.java
mraible: @arvindkrishnakumar-okta It seems like we could create the `DefaultServerOAuth2AuthorizationRequestResolver` before passing it to `Okta`'s static methods, but that would change the method signature for `configureOAuth2WithPkce()`. I believe that would be a breaking change and require a major version bump. Do you have any other ideas?

arvindkrishnakumar-okta: @mraible I think that is the only way out here to resolve the clobbering.

oauth2/src/main/java/com/okta/spring/boot/oauth/Okta.java
arvindkrishnakumar-okta: I guess it should not be a problem sending it for Okta too.

oauth2/src/main/java/com/okta/spring/boot/oauth/ReactiveOktaOAuth2AutoConfig.java
deepu105: Would it be possible to overload the method with new signature and deprecate the current one?

PR #604, opened 2023-09-27
OKTA-634004: Add Snyk software dependency scan

arvindkrishnakumar-okta:


arvindkrishnakumar-okta:
Build is failing presently due to the fact that snyk runs are restricted to `main` branches only. We should be good with this PR and once we merge the sync scans should work.

bretterer (approved):

PR #605, opened 2023-09-27
Add JDK image to Snyk Job

arvindkrishnakumar-okta:


arvindkrishnakumar-okta:
Build is failing presently due to the fact that snyk runs are restricted to main branches only. We should be good with this PR and once we merge the sync scans should work.

bretterer:
> Build is failing presently due to the fact that snyk runs are restricted to main branches only. We should be good with this PR and once we merge the sync scans should work.

You can add `run-on-non-main` to the config and make it run on all branches

bretterer (approved):

.circleci/config.yml
```diff
@@ -26,13 +26,16 @@ jobs:
...
+      - image: cimg/openjdk:17.0.6
     steps:
       - attach_workspace:
           at: ~/project
+      - run: rm -rf mvnw mvnw.cmd .mvn/wrapper
       - general-platform-helpers/step-load-dependencies
       - general-platform-helpers/step-run-snyk-monitor:
           scan-all-projects: true
           skip-unresolved: false
+          run-on-non-main: true
```
itaimarongwe-okta, line 38: Make sure to remove this after this testing is done.

arvindkrishnakumar-okta: Removing this would mean all PRs running on feature branches are bound to fail because the job `snyk-scan` will error out. cc: @bretterer

itaimarongwe-okta: @aravindanranganathan-okta hmmm, you're correct. Apologies,  this shouldn't fail if set to false or not be present. It should simply skip the scan - I'll chase this down.  Feel free to leave it.

PR #666, opened 2024-04-02
Enforce `nimus-jose-jwt` `9.37.3` in pom.xml

arvindkrishnakumar-okta:
OKTA-699955

https://mvnrepository.com/artifact/org.springframework.security/spring-security-oauth2-jose/6.2.3 pulls in an older version of nimbus-jose-jwt, so we are enforcing a newer `9.37.3` with this PR.


bryanapellanes-okta (approved):
👍

PR #752, opened 2024-12-22
Fix usage of OktaJwtAuthenticationConverter to use constructor with props

mpanik:
Updated the autoconfiguration to use the OktaJwtAuthenticationConverter constructor that accepts OktaOAuth2Properties instead of the one that accepts a group name. This change allows the authoritiesClaimName property to be used for mapping authorities from the "permissions" claim, ensuring proper configuration and functionality.


arvindkrishnakumar-okta:
@mpanik Thanks for the PR!

If not already, can you submit a [CLA](https://developer.okta.com/cla/)?

mpanik:
@arvindkrishnakumar-okta signed and sent.

